PR #2, opened 2018-09-19
Fix integer overflow bug in binary search

crazymerlyn:
`left + right` can be outside the range of 32bit numbers even if `left` and `right` are not. This can lead to bugs on 32bit systems. See [Nearly All Binary Searches and Mergesorts are Broken](https://ai.googleblog.com/2006/06/extra-extra-read-all-about-it-nearly.html).


AnshulMalik:
Great catch there! @christianbender . Thanks :)

PR #5, opened 2018-10-01
Changed the code to more idiomatic Rust.

bofh69:
Fixed most of clippy's warnings.
The linear search is more idiomatically solved by an iterator, but then the algorithm is harder to follow.


AnshulMalik:
Thanks @bofh69

Issue #3, opened 2018-09-19
Add more algorithms in Rust Language


r0hit-gupta:
This issue is fairly easy and most beginners should be comfortable implementing algorithms of their choice in Rust. In case of any problem, raise an issue or just discuss it below.

**Graphs**
 - [ ] DFS

**Dynamic Programming**
 - [ ] Longest increasing subsequence()
 - [ ] Rod cut
 - [ ] Egg Dropping Puzzle
 
**Data Structures**
 - [ ] Graph
   - [ ] Directed
   - [ ] Undirected
 - [ ] Trie
 - [ ] AVL Tree

**General**
 - [ ] N-Queens
 - [ ] Graph Coloring
 - [ ] Run Length Encoding

**Project Euler**, **LeetCode**, etc.
 - Currently, we don't accept problems from those and similar websites

nlynchjo:
@r0hit-gupta I am also looking for a first time contribution! Mind if I take on some of the data structures?

r0hit-gupta:
@0xazure @nlynchjo Please feel free to contribute and work on anything you like. Take an idea from other repositories if you do not know how to begin. Send a pull request and if there is something not right, we will let you know. Cheers!

darkyshiny:
@r0hit-gupta I'll take a crack at the KMP string pattern matching 👍

bofh69:
@0xazure - sorry, I didn't read your comment until now. I just added a PR with insertion sort.

0xazure:
@bofh69 thanks for letting me know! I hadn't actually started implementing it yet, so I'll focus on the selection sort. I'm definitely interested in how you did it in Rust, so I'll have a look at your PR!

bofh69:
@0xazure Great! I felt a bit stupid not checking the comments here first...
My solution is probably not the best, I'm still learning Rust, but now it is decent.

eisterman:
I'm actually working on Merge Sort

AnshulMalik:
I think there is already a PR for Merge Sort @eisterman

Jay9596:
I'd like to work on implementing Queue, if no one has claimed it.

AnshulMalik:
Go ahead @Jay9596 :)

ra-kete:
I'd like to try my hands on heap sort.

DanielSauve:
I'll take a shot at making a Stack

BaxterEaves:
[...]

PR #7, opened 2018-10-01
added shell sort

sn99:
#3 

I have added 2 ways of shell sort the one that performs better in benchmarks shall be final


efugier:
"posicion" or "insercion" are not english words, please refactor your code !
Also, add your functions to the `mod.rs` file and put the tests in `lib.rs`.

sn99:
Will make new pull request

PR #9, opened 2018-10-02
add knapsack bottom up solution

AnshulMalik:
Fixes part of #3


AnshulMalik:
Umm, knapsack is a problem which works on numbers I think, what do you suggest?

efugier:
Well for instance, it this implementation doesn't work on `u64`, maybe try to make it generic for all number types ?

Issue #8, opened 2018-10-02
Add README.md


AnshulMalik:
We should have a readme with some info just like other repositories have [Python](https://github.com/TheAlgorithms/Python). 

It should just highlight the things that we have implemented and the things which are yet to be implemented.

For now, it should just contain the headlines for different categories of algorithms and later, we can fill them with specific algorithms.

camilosampedro:
I want to work on this issue. Can I help you?

AnshulMalik:
Yes, sure @camilosampedro. Thanks :)

PR #10, opened 2018-10-02
A new README.md

camilosampedro:
This Pull Request adresses #8 

It is based on the Python repository README.md. I specially looked for Python-specific references that could be changed to Rust in the document, and I consider that this is a very general algorithms README.


AnshulMalik:
Hey, we are going to have separate readme for the modules, so the sorting docs will go to sorting readme. 
So, we should just keep the headings here and the algorithms implemented and to be implemented.

r0hit-gupta:
@AnshulMalik that is right. Let us have different README files for every different category. Can you please move these files to their appropriate directories? Thanks.

camilosampedro:
I just updated it, separating it into different README.md files, in each of the directories.

AnshulMalik:
We need to update the status of implemented algorithms as we progress.
Thanks @camilosampedro

PR #11, opened 2018-10-02
Selection Sort

guptasidharth1:
Time Complexity: O(n2) as there are two nested loops.
Auxiliary Space: O(1)
The good thing about selection sort is it never makes more than O(n) swaps and can be useful when memory write is a costly operation.

### DIFF
--- a/src/sorting/selection-sort.rs
+++ b/src/sorting/selection-sort.rs
@@ -1,0 +1,20 @@
+fn selection_sort(array: &mut [i32]) {
+    let mut min;
+    for i in 0..array.len() {
+        min = i;
+        for j in (i+1)..array.len() {
+            if array[j] < array[min] {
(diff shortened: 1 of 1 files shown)


AnshulMalik:
Looks good @guptasidharth1 just one thought though.

Currently, it works just for numbers, It'd be complete if we make it generic, see https://github.com/TheAlgorithms/Rust/blob/master/src/sorting/bubble_sort.rs if you need some reference.

AnshulMalik:
We can have tests in place of the main function, since we are not going to use the main function anyway.

efugier:
Hi, please read our [CONTRIBUTING.md](https://github.com/TheAlgorithms/Rust/blob/master/CONTRIBUTING.md) and make changes accordingly.

PR #16, opened 2018-10-02
Create Longest-Increasing-Subsequence.rs

guptasidharth1:
The Longest Increasing Subsequence (LIS) problem is to find the length of the longest subsequence of a given sequence such that all elements of the subsequence are sorted in increasing order.


yashwanthreddyg:
@guptasidharth1 Please try to keep variable names as descriptive as possible. Also, add comments where you think they will help the reader. The repo should be approachable to new-comers not only in Rust but also CS in general. :)

AnshulMalik:
It'll be a lot more helpful if people contribute their original work to open source projects.

https://codereview.stackexchange.com/questions/187337/longest-increasing-subsequence-algorithm

PR #17, opened 2018-10-02
Merge-sort

chronos25:
broke down method into two parts - one that determines from which side to pull, and one that actually advances that iterator.

**Merge sort** is a sorting technique based on **divide and conquer technique**. With worst-case time complexity being **Ο(n log n)**, it is one of the most respected algorithms. Merge sort first divides the array into equal halves and then combines them in a sorted manner.

### DIFF
--- a/src/sorting/merge-sort.rs
+++ b/src/sorting/merge-sort.rs
@@ -1,0 +1,58 @@
+use std::iter::Peekable;
+use std::cmp::Ordering;
+
+struct MergeAscending<L, R>
+where
+    L: Iterator<Item = R::Item>,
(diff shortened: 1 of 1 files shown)


yashwanthreddyg:
Hey @chronos25. Shouldn't we have a common merge_sort function which internally uses MergeAscending [or Descending]?

AnshulMalik:
I don't suppose that's you.

https://users.rust-lang.org/t/solved-merge-multiple-sorted-vectors-using-iterators/6543

Please contribute your original work as your contribution, and not steal.

PR #24, opened 2018-10-02
Add counting sort

efugier:
Added regular and generic counting sort, as well as the tests

### DIFF
--- a/src/sorting/counting_sort.rs
+++ b/src/sorting/counting_sort.rs
@@ -1,0 +1,80 @@
+/// In place counting sort for collections of u32
+/// O(n + maxval) in time, where maxval is the biggest value an input can possibly take
+/// O(maxval) in memory
+/// u32 is chosen arbitrarly, a counting sort probably should'nt be used on data that requires bigger types.
+pub fn counting_sort(arr: &mut [u32], maxval: usize) {
+    let mut occurences: Vec<usize> = vec![0; maxval + 1];
+
+    for &data in arr.iter() {
+        occurences[data as usize] += 1;
+    }
+
+    let mut i = 0;
+    for (data, &number) in occurences.iter().enumerate() {
+        for _ in 0..number {
+            arr[i] = data as u32;
+            i += 1;
+        }
+    }
+}
+
+use std::ops::AddAssign;
+/// Generic implementation of a counting sort for all usigned types
+pub fn generic_counting_sort<T: Into<u64> + From<u8> + AddAssign + Copy>(
+    arr: &mut [T],
+    maxval: usize,
+) {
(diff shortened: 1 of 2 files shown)


AnshulMalik:
Nice @efugier, tests would look good in the `counting_sort.ts` :)

src/sorting/counting_sort.rs
```diff
@@ -0,0 +1,44 @@
+/// In place counting sort for collections of u32
+/// O(n + maxval) in time, where maxval is the biggest value an input can possibly take
+/// O(maxval) in memory
+/// u32 is chosen arbitrarly, a counting sort probably should'nt be used on data that requires bigger types.
+pub fn counting_sort(arr: &mut [u32], maxval: usize) {
+    let mut occurences: Vec<usize> = vec![0; maxval + 1];
+
+    for &data in arr.iter() {
+        occurences[data as usize] += 1;
```
AnshulMalik: I was just thinking that we can we use counting sort, if we have negative numbers in the array?

efugier:
Have updated the branch, ready to merge =)

By the way, I'd love to become a collaborator on this organization :upside_down_face:

src/sorting/counting_sort.rs
```diff
@@ -0,0 +1,80 @@
...
+    let mut i = 0;
+    // current data point, necessary to be type-safe
+    let mut data = T::from(0);
+
+    for &number in occurences.iter() {
+        for _ in 0..number {
+            arr[i] = data;
+            i += 1;
+        }
+        data += T::from(1);
```
AnshulMalik: What's this line(40) for?

efugier: Well, I had great trouble implementing the generic version of the test because you have to use the input type as indexes in an array.
Then you have to convert array indexes into the input type in order to re-fill the collection. That's the tricky part.

Array indexes are usize, but the only thing that can be converted from usize is... usize itself, so the _generic_ version could only work on usize with the standard implementation, not that generic, you'll agree.

So I imagined this trick to keep the  genericity: `data` is simply a counter that represents the current index occurence **as the input type**.

`T::from(1)` uses the `From<u8>` trait bound to convert 1 (treated as a `u8`) into `T`.

the requirements for `T` are:
1. Something that can be converted to `u64`
2. Something that can be converted from `u8`

So this implementations works with all the unsigned numbers, which was the targeted goal for a **safe** generic counting sort.

efugier:
Why add printing in the test?

AnshulMalik:
Oops my bad, will fix it soon. Thanks to your eagle eyes :)

efugier:
No problem!
Also remove the debug bound!

PR #28, opened 2018-10-03
Add Merge Sort (recursive) with relative tests

eisterman:


efugier:
Please run `cargo fmt`  before commiting.
Even though I agree that the tests should be next to the functions, the current paradigm is to put them in `lib.rs` so please do so.

AnshulMalik:
Sorry, Closing due to inactivity.

PR #45, opened 2018-10-04
Add Circular Queue implementation

Jay9596:
- Added a Generic Circular Queue implementation with basic operations.
- Added tests for operations


efugier:
Nice one, please use slices instead of vects

Jay9596:
How and where do you want me to use slices here ?

efugier:
Woops, misread something, forget about it sorry !

src/data_structure/circular_queue.rs
```diff
@@ -0,0 +1,105 @@
...
+            index: 0,
+            front: 0,
+            capacity,
+        }
+    }
+
+    // Adds the element to the circular queue
+    // Once capacity is reached, new items will overwrite old ones.
+    pub fn enqueue(&mut self, e: T) {
+        if self.len() < self.capacity() {
```
AnshulMalik: We can save a call here, since we can directly use `self.capacity` :)

src/data_structure/circular_queue.rs
```diff
@@ -0,0 +1,105 @@
...
+        self.index = (self.index + 1) % self.capacity();
+    }
+
+    // Returns an Option with first element of the list,
+    // if it is empty returns None
+    pub fn dequeue(&mut self) -> Option<T> {
+        if self.is_empty() {
+            return None;
+        }
+        let ele = self.data.swap_remove(self.front as usize);
```
AnshulMalik: I am not sure if we want to do this, since `swap_remove` removes the element at `self.front` and replaces it with the last element of the array, so I think it places the element at the end of queue to the front, and If after this, we `enqueue` other elements, they will probably be out-of-order.
We should also add more tests to cover various scenarios possible :)

src/data_structure/circular_queue.rs
```diff
@@ -0,0 +1,105 @@
...
+        assert!(queue.dequeue().is_none());
+        queue.enqueue(1);
+        assert!(!queue.is_empty());
+        queue.enqueue(2);
+        assert_eq!(queue.len(), 2);
+        queue.dequeue();
+        queue.dequeue();
+        assert!(queue.is_empty());
+    }
+}
```
AnshulMalik: We can have few more tests to test whether the elements returned from dequeue are the ones we want them to

Jay9596: I'll add the test to check for the dequeue elements.

src/data_structure/circular_queue.rs
Jay9596: This is necessary as pushing into a newly created vec with capacity will give error. 
This checks helps to check whether we have/had an element at that position, or we need to push a new element.

src/data_structure/circular_queue.rs
Jay9596: `enqueue` always places the element, so it's never out of bounds. It will overwrite the element at the front **always**. To have O(1) operatoins in a vec, this is what I used. If you still want the change, I'll have to rewrite it in VecDeque.

AnshulMalik: Okay, can you check for this test is it passes?

```
let mut queue: Queue<i32> = Queue::with_capacity(5);
queue.enqueue(1);
queue.enqueue(2);
queue.enqueue(3);
queue.dequeue(); // this should remove 1
assert_eq!(queue.dequque().unwrap(), 2)

```

Jay9596: Yes, I'm working on adding test for the dequeue elements.

Jay9596:
I've added the new tests for dequeue, please verify.

src/data_structure/circular_queue.rs
AnshulMalik: I was just saying if we can do `self.capacity` in place of `self.capacity()` since the function return this same thing.

src/data_structure/circular_queue.rs
AnshulMalik: I am still thinking how is it working, can you try this?
```
        let mut queue: Queue<i32> = Queue::with_capacity(5);
        assert!(queue.dequeue().is_none());
        queue.enqueue(1);
        queue.enqueue(2);
        queue.enqueue(3);
        queue.enqueue(4);
        assert_eq!(queue.dequeue(), Some(1));
        assert_eq!(queue.dequeue(), Some(2));
        assert_eq!(queue.dequeue(), Some(3));
        assert_eq!(queue.dequeue(), Some(4));
```

src/data_structure/circular_queue.rs
Jay9596: OK, got it. Instead of a function call, use the value from `self` directly. 
Sorry misunderstanding on my part. 😞

src/data_structure/circular_queue.rs
Jay9596: This fails, it fails at `Some(3)`, due to the new vec having a different length the front is miscalculated.
What do you suggest I do ?

AnshulMalik: We can chat about it on Gitter :)

Volpe95:
Closing this pull request as not all checks are satisfied and inactivity.

PR #47, opened 2018-10-04
added radix sort

sn99:
#3

### DIFF
--- a/src/sorting/radix_sort.rs
+++ b/src/sorting/radix_sort.rs
@@ -1,13 +1,13 @@
-pub fn radix_sort(data: &mut [isize]) {
+pub fn faster_radix_sort(data: &mut [isize]) {
     // radix is 256 for faster sort
     let mut buckets = vec![vec![]; 256];
 
-    for n in 0..4 {
+    for z in 0..4 {
         for r in data as &[isize] {
             let val = *r;
             // get the nth byte which has 256 possible values
-            let mut which = (val >> (8 * n)) & 0xFFisize; // It's a bit like (val / (256 to the power of n)) % 256
-            if n == 3 {
+            let mut which = (val >> (8 * z)) & 0xFFisize; // It's a bit like (val / (256 to the power of n)) % 256
+            if z == 3 {
                 //  The XOR instruction (^) is to flip the sign bit so that negative values work properly
                 which ^= 0x80;
             }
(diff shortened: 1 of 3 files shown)


AnshulMalik:
Hey @sn99 this is nice work! Can we use the existing `counting_sort` implementation that we have in here? Since it'd be more intuitive to understand.

sn99:
@AnshulMalik I don't quite understand you ?

src/sorting/radix_sort.rs
```diff
@@ -0,0 +1,52 @@
+pub fn radix_sort(data: &mut [isize]) {
+    let mut buckets = vec![vec![]; 256];
+
+    for n in 0..4 {
+        for r in data as &[isize] {
+            let val = *r;
+            let mut which = (val >> (8 * n)) & 0xFFisize;
```
AnshulMalik: this part is a bit hard to understand, we can use some comment here or try an easier version :)

AnshulMalik:
Let's have a conversation in Gitter channel :)

sn99:
@AnshulMalik I don't think we can change `radix_sort.rs` to use `counting_sort.rs` functions as we use a completely modified version of *counting sort* for radix sort :)

sn99:
@efugier I don't think sorting algo's work that way, a sorting algo must be complete in itself and simply take a *slice* for sorting

efugier:
They do. What if you want to sort a huge array of 0 and 1, do you really want to walk through your array for nothing?

sn99:
@efugier let's talk on gitter :)

efugier:
You can't pass it negative integers, the type checking wouldn't allow it.

Don't forget that this repo is a compilation of algorithms, and the basic counting sort does not take negative integers nor does it checks for the max and min value. 

A first pass to determine the range of the data would result in a x2 when sorting know ranges, such as bits array, which are the major use cases for this algorithm: very small (known) range but huge array.

So no, the first pass does not belong in here.
Plus, calculating the max is really easy and would just add noise in a more complex thing. 

(Also such discussions belongs here, in public!)

AnshulMalik:
Both of you have got valid points to make here 👍 ,

Let's do a research on this and go back in past to see how it was originally intended to work and we'll implement it that way, 
And we can also use our innovative ideas to improve it in another implementation. But I agree with @efugier that we should have one implementation which is original.

efugier:
I think it's best to revert it :)

As a side note, `my_array.iter().max()` will cleanly and idiomatically give you the maximum if needed !

sn99:
Yeah, I know about `.max()` in applied on iterators but a method that shows how max can be found is better for understanding :)

I have reverted the changes made in `counting_sort.rs` :)

PR #49, opened 2018-10-07
Add k-means under general algorithms

BaxterEaves:


AnshulMalik:
This is really great! @BaxterEaves Thanks so much for helping us.
Just one thought, do you think we should also support `f32` as well? I think we should :)

AnshulMalik:
And, we should also add tests for higher dimensions data.

BaxterEaves:
If the goal of the repo is to demonstrate algorithms I'm unclear about the benefit of supporting `f32`. That aside how would you prefer the `kmeans` function to be generalized to support both `f32` and `f64`? Normally, I would bring in the `num` crate for it's `Float` trait and write `kmeans<T: Float>`.

AnshulMalik:
@BaxterEaves you are right, goal is to demonstrate algorithms, but it's also to write them such that they are usable by a broader audience. So people might use the same algorithms for different purpose on different types of data. What do you think @efugier ?

BaxterEaves:
Because I thought you might not want to being in external dependences, I used macros to implement `kmeans::f32::kmeans` and `kmeans::f64::kmeans` which are re-imported as `kmeans_f32` and `kmeans_f64` , respectively.

AnshulMalik:
This is really great! Thanks :)

PR #53, opened 2018-10-24
Mergesort

onsah:
I tried to write a merge sort implementation. It looks like it works from the tests.

### DIFF
--- a/src/sorting/merge_sort.rs
+++ b/src/sorting/merge_sort.rs
@@ -1,0 +1,64 @@
+pub fn merge_sort<T: Ord + Clone>(arr: &mut [T]) {
+    let mut mem = arr.to_vec();
+    _merge_sort(arr, &mut mem);
+}
+
+fn _merge_sort<T: Ord + Clone>(arr: &mut [T], mem: &mut [T]) {
+    assert_eq!(arr.len(), mem.len());
+    match arr.len() {
+        0...1 => (),
+        len => {
+            let (left_mem, right_mem) = mem.split_at_mut(len / 2);
+            {
+                let (left, right) = arr.split_at_mut(len / 2);
+                _merge_sort(left_mem, left);
+                _merge_sort(right_mem, right);
+            }
+            merge(arr, left_mem, right_mem);
+        }
+    }
+}
+
+fn merge<T: Ord + Clone>(dest: &mut [T], left: &[T], right: &[T]) {
+    let mut lit = left.iter().peekable();
+    let mut rit = right.iter().peekable();
+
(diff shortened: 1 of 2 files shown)


plus7wist:
I like this code.

However, I think there's some points that can be better.

1. `...` range patterns are deprecated, we can use `..=` instead.
2. This code requires sorted type `T` impl `Clone`, this is not really good for a sorting algorihtm. Is there a way to remove that (in safe rust)?

AnshulMalik:
@plus7wist Thanks for your taking your time to review this :)
We can certainly use your help at other PRs as well.

I agree with point 1.
But how merge sort works is by creating another copy of the original array. That's how mergesort is defined, so I think we need the `Clone` anyway. Let me know your thoughts on this.
Thanks

plus7wist:
Of course, we have to use a buffer in O(n) size to achieve our target. But this buffer need not to be a array (or vec) of `T`, we can imagine such a  implementation: In devide step, we move
 objects in our array to a new buffer, and in merge step, we move them back.

In such a way, we don't call `Clone::clone()` on objects in array. So our sort function can be used to sort something can not be cloned, such as complex structs contains a `Mutex`.

And, if the user *impl*ed a `Clone` by hand, with some nontrivial actions, such like object counting, our sort function will not aggravates him.

---

If we use unsafe rust, just create a uninitialized buffer, our target might be easy, but in safe rust, I don't have a good idea now.

---

I am sorry I didn't make my meaning clear at the first time... English is not my mother tongue.

plus7wist:
Well, I have made a ugly version of my proposition. I sort the index of the array instead of the array itself, then use `Slice::swap` to re-arrange array to avoid `clone`. Here's my code:

<https://github.com/plus7wist/Rust/blob/master/src/sorting/merge_sort.rs>

There must be a better solution.

plus7wist:
@AnshulMalik Can we use a buffer typed `Vec<&T>` to do this?

AnshulMalik:
Yeah, I can not see it, we can just move the items back and forth, instead of cloning them.

plus7wist:
@AnshulMalik I find that the current stable sort implemention in Rust std is just merge sort! And that don't requires `T: Clone`.

<https://doc.rust-lang.org/std/primitive.slice.html#current-implementation-6>

I'd like to view the code to learn that.

plus7wist:
Well, generally speaking, the std code use unsafe operations to do something like "move", but they just do shadow copy instead of deep clone, so that sorting don't need any ownership of `T`.

I think we have collected all keys now.

PR #58, opened 2018-10-28
Add Binary Tree

elpiel:
Add binary tree for #3

### DIFF
--- a/src/data_structure/binary_tree.rs
+++ b/src/data_structure/binary_tree.rs
@@ -1,0 +1,66 @@
+#[derive(PartialEq)]
+pub struct Node<'a> {
+    val: &'a str,
+    left: Option<Box<Node<'a>>>,
+    right: Option<Box<Node<'a>>>,
+}
+
+impl<'a> Node<'a> {
+    pub fn insert(&mut self, new_val: &'a str) {
+        if self.val == new_val {
+            return;
+        }
+        let target_node = match new_val < self.val {
+            true => &mut self.left,
+            false => &mut self.right,
+        };
+        match target_node {
+            &mut Some(ref mut sub_node) => sub_node.insert(new_val),
+            &mut None => {
+                let new_node = Node {
+                    val: new_val,
+                    left: None,
+                    right: None,
+                };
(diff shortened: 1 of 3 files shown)


src/data_structure/binary_tree.rs
```diff
@@ -0,0 +1,66 @@
+#[derive(PartialEq)]
+pub struct Node<'a> {
+    val: &'a str,
```
AnshulMalik: Let's make this generic so that can can have anything as a value

AnshulMalik: Do you think it's better to name it as `BinaryNode` since it has at most two children?

src/data_structure/binary_tree.rs
```diff
@@ -0,0 +1,66 @@
...
+    left: Option<Box<Node<'a>>>,
+    right: Option<Box<Node<'a>>>,
+}
+
+impl<'a> Node<'a> {
+    pub fn insert(&mut self, new_val: &'a str) {
+        if self.val == new_val {
```
AnshulMalik: Why are we not inserting if the incoming value matches the current?
I think you are trying to implement a `Binary Search Tree`

src/data_structure/binary_tree.rs
```diff
@@ -0,0 +1,66 @@
+#[derive(PartialEq)]
+pub struct Node<'a> {
+    val: &'a str,
+    left: Option<Box<Node<'a>>>,
+    right: Option<Box<Node<'a>>>,
+}
+
+impl<'a> Node<'a> {
```
AnshulMalik: I was thinking of having different types of nodes, like for a Binary Tree, for B-Tree.

And then different traits for different kinds of trees.
Like for `BinaryTree`, the trait will have an `insert` method, which is simple, just insert wherever you want.
But for `BinarySearchTree` the trait's method will impose this condition of `Order` between among children.

elpiel:
I will try to take a look at the comments and see what I can do later.

Shub1427:
Binary Search Tree has been implemented and merged in https://github.com/TheAlgorithms/Rust/pull/76

PR #63, opened 2018-12-24
added the already implemented selection sort to documentation

KappaDistributive:
I've added an entry for Selection Sort to the documentation and fixed the alphabetical arrangement of all sorting algorithms.

Unfortunately, there was no picture for Selection Sort available on Wikipedia and no visualisation available on Toptal, so these spots remain empty.


AnshulMalik:
Thanks @KappaDistributive

PR #65, opened 2018-12-25
a rudimentary implementation of a weighted graph

KappaDistributive:
This is a rudimentary implementation of a weighted graph.

It has been written with the intention to implement Dijkstra in a future PR.

### DIFF
--- a/src/data_structure/graph.rs
+++ b/src/data_structure/graph.rs
@@ -17,7 +17,7 @@ impl<N, E, Ty> Graph<N, E, Ty>
 where
     Ty: EdgeType,
 {
-    fn new() -> Self {
+    pub fn new() -> Self {
         Graph {
             nodes: HashMap::new(),
             edges: HashMap::new(),
@@ -26,12 +26,12 @@ where
     }
 
     /// Returns true iff graph is directed.
-    fn is_directed(&self) -> bool {
+    pub fn is_directed(&self) -> bool {
         Ty::is_directed()
     }
 
     /// Add node to graph and return its assigned index.
-    fn add_node(&mut self, weight: N) -> usize {
+    pub fn add_node(&mut self, weight: N) -> usize {
         let node = Node { weight: weight };
         let mut index = self.nodes.len();
 
@@ -49,7 +49,7 @@ where
 
     /// Remove node indexed at 'index' and all edges containing that
     /// node. Optionally returns the weight of the removed node.
-    fn remove_node(&mut self, index: usize) -> Option<N> {
+    pub fn remove_node(&mut self, index: usize) -> Option<N> {
         match self.nodes.remove(&index) {
             Some(node) => {
                 // Only keep those edges that don't contain index as head or tail
@@ -66,7 +66,7 @@ where
     /// starting node has index head and whose ending node has index
     /// tail to the graph. If this succeeds, return index assigned to
     /// that edge. Otherwise return None.
-    fn add_edge(&mut self, weight: E, head: usize, tail: usize) -> Option<usize> {
+    pub fn add_edge(&mut self, weight: E, head: usize, tail: usize) -> Option<usize> {
         if !self.nodes.contains_key(&head) || !self.nodes.contains_key(&head) {
             None
         } else {
@@ -94,14 +94,14 @@ where
     }
 
     /// Remove edge indexed at 'index' 
-    fn remove_edge(&mut self, index: usize) {
+    pub fn remove_edge(&mut self, index: usize) {
         self.edges.remove(&index);
     }
 
     /// Return first index of an edge of the form head --> tail.  This
     /// is not necessarily the unique index since we explicitly allow
     /// multiple edges between two nodes.
-    fn find_edge(&self, head: usize, tail: usize) -> Option<usize> {
+    pub fn find_edge(&self, head: usize, tail: usize) -> Option<usize> {
         let edges: Vec<usize> = self.find_n_edges(1,head,tail);
         if edges.len() == 0 {
             return None;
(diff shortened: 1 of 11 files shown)


src/data_structure/graph.rs
```diff
@@ -0,0 +1,221 @@
...
+    /// Returns true iff graph is directed.
+    fn is_directed(&self) -> bool {
+        Ty::is_directed()
+    }
+
+    /// Add node to graph and return its assigned index.
+    fn add_node(&mut self, weight: N) -> usize {
+        let node = Node { weight: weight };
+        let mut index = 0usize;
+        while self.nodes.contains_key(&index) {
```
AnshulMalik: This while loops would be inefficient here. Can we do something like `self.nodes.insert(self.nodes.len(), node)`?

KappaDistributive:
I am new to Rust and forgot to mark the module as public. This should be fixed now to Travis CI didn't seem to take that change into consideration. 

Any insight is appreciated.

src/data_structure/graph.rs
```diff
@@ -0,0 +1,221 @@
...
+    /// Return first index of an edge of the form head --> tail.  This
+    /// is not necessarily the unique index since we explicitly allow
+    /// multiple edges between two nodes.
+    fn find_edge(&self, head: usize, tail: usize) -> Option<usize> {
+        for (index, edge) in self.edges.iter() {
+            // The latter part of the following disjunction is needed
+            // to identify [head,tail] with [tail,head] in undirected
+            // graphs.
+            if [edge.head, edge.tail] == [head, tail]
+                || (self.is_directed() && [edge.head, edge.tail] == [tail, head])
```
AnshulMalik: This code look duplicate as in `find_edges`, can we reuse some of this?

AnshulMalik:
I was thinking, we should also provide a method that can expose iterable over edges and nodes. That way we can use this for arbitrary things.

src/data_structure/graph.rs
KappaDistributive: Yes, thanks for the suggestion. As far as I can tell there is no reason to assign the smallest available index to a new node or edge.

src/data_structure/graph.rs
KappaDistributive: You're right. I'll get rid of that duplicate code.

src/data_structure/graph.rs
KappaDistributive: Ive thought about it and
`self.nodes.insert(self.nodes.len(), node)`
will not work if we delete and insert a bunch of nodes (i.e. it will result in duplicate usages of a certain index). But it still is the most reasonable first attempt for an index and by decreasing it, if necessary, we will always find an available index, i.e. 
`let mut index = self.nodes.len();
while self.nodes.contains_key(&index) {
            index -= 1;
        }`
will always result in an available index and, most of the time, it will do so in one step.

KappaDistributive:
I haven't implemented iterators in this last commit, yet. While I intend to add those features, I'm not sure how to best approach this and hence need to do some research first.

If someone is willing to offer some advice, I'd appreciate it!

KappaDistributive:
> You should try making the functions public as well.

Thanks. I had no idea!

src/data_structure/graph.rs
plus7wist: I think we can use something like "Queue" or "Stack" to maintain a list of avaliable indices. Every time a node/edge was deleted, add the delected index to that free list. When we want to add a item, we pop one index from the free list. If that list is empy, then we use `self.xxx.len()` as the index.

I have write some code: <https://github.com/plus7wist/Rust/blob/d5039cee609df8d2823ddab01df40fa22febb2a6/src/data_structures/graph.rs#L280>

plus7wist: > The indices, in my mind, are purely for bookkeeping purposes. If a user wants to add labels to their nodes (edges), they should add them to their weights. So, a weighted graph with labels may have weights of type (String, i32), for example.
> 
> If this agrees with you, I'd suggest renaming 'weight' into 'data' or 'payload' to reflect this more general purpose.
> 
> But I'm open to suggestions on this.

Well, if we add our label into `weight`, we can't find a node/edge by our label effectively. In this problem, I agree with @AnshulMalik more.

src/data_structure/graph.rs
```diff
@@ -0,0 +1,294 @@
...
+            }
+        }
+    }
+
+    /// Given two indices head and tail, attempt to add an edge whose
+    /// starting node has index head and whose ending node has index
+    /// tail to the graph. If this succeeds, return index assigned to
+    /// that edge. Otherwise return None.
+    pub fn add_edge(&mut self, weight: E, head: usize, tail: usize) -> Option<usize> {
+        if !self.nodes.contains_key(&head) || !self.nodes.contains_key(&head) {
```
plus7wist: Is there a typo? I think the second `head` might be `tail`.

src/data_structure/graph.rs
```diff
@@ -0,0 +1,294 @@
...
+        self.find_n_edges(0, head, tail)
+    }
+
+    /// Returns up to the first n edges of the form head --> tail. If
+    /// n == 0, it returns all available edges.
+    pub fn find_n_edges(&self, n: usize, head: usize, tail: usize) -> Vec<usize> {
+        let mut result: Vec<usize> = Vec::new();
+        for (index, edge) in self.edges.iter() {
+            if [edge.head, edge.tail] == [head, tail]
+                || (self.is_directed() && [edge.head, edge.tail] == [tail, head])
```
plus7wist: I don't understand this line.

I think when our graph is *not* directed, edge from `tail` to `head` can be treated as edge from `head` to `tail`.

AnshulMalik: Yea agree @plus7wist this looks little complicated.

AnshulMalik:
Closing due to inactivity, but feel free to open it again :)

PR #68, opened 2019-02-09
Separate tests more and add another test

arnavb:
This way, it's easy to determine from the name of the test which ones failed and which ones didn't. Many feel that a test should only have one focus. I can do this for other files as well.


AnshulMalik:
Looks good this way. Thanks :)

PR #71, opened 2019-02-15
Separate all tests in searching folder and some others

arnavb:
Self explanatory. Basically like what I did in #68. Add some tests as well.

### DIFF
--- a/src/sorting/heap_sort.rs
+++ b/src/sorting/heap_sort.rs
@@ -122,4 +122,25 @@ mod tests {
         heap_sort(&mut arr);
         assert_eq!(&arr, &[1, 2, 3, 4]);
     }
+
+    #[test]
+    fn odd_number_of_elements() {
+        let mut arr = vec![3, 4, 2, 1, 7];
+        heap_sort(&mut arr);
+        assert_eq!(&arr, &[1, 2, 3, 4, 7]);
+    }
+
+    #[test]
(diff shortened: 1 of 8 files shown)


arnavb:
@AnshulMalik I've implemented/separated more tests and this should be ready to merge once CI passes.

src/sorting/heap_sort.rs
```diff
@@ -122,4 +122,32 @@ mod tests {
...
+
+    #[test]
+    fn unusual_element_at_end() {
+        let mut arr = vec![3, 4, 2, 1, 7, 33333];
+        heap_sort(&mut arr);
+        assert_eq!(&arr, &[1, 2, 3, 4, 7, 33333]);
+    }
+
+    #[test]
+    fn unusual_element_at_start() {
```
AnshulMalik: Do we need these unusual tests?

arnavb: @AnshulMalik I don't know if we _need_ them, but I've learned that when testing, unusual end cases should definitely be considered.

arnavb: However, if you feel that these tests are unnecessary, I can remove them.

AnshulMalik: Yep, feels like redundant to me. Better if we keep them concise :)

PR #76, opened 2019-07-15
Binary Search Tree

JarredAllen:
I coded a simple Binary Search Tree.

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,7 +61,7 @@ These are for demonstration purposes only.
 ## [Search Algorithms](./src/searching)
 
 - [Linear](./src/searching/linear_search.rs)
-- [Binary](./src/seraching/binary_search.rs)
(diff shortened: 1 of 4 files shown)


AnshulMalik:
Just need to run cargo fmt

JarredAllen:
I have already run cargo fmt on this, and it gave a different output from what the one on the test does.

If it's relevant, I have rustfmt version 0.10.0, which is what it installed when I ran `cargo install rustfmt`, and I just ran `cargo fmt -- --force` to format it.

AnshulMalik:
This is what the test script use: `rustup component add rustfmt-preview` may be we can try this once.

JarredAllen:
I installed a newer version of rustfmt through rustup and ran it on the code and it made some changes from the older version. Apparently, installing it through cargo didn't get me the latest version. But it looks like everything is good now.

PR #84, opened 2019-09-11
"Merge Sort" without Clone

plus7wist:
Add a generic "Merge Sort" implemention.

This `merge_sort<T>(&mut [T])` do not requires `T: Clone`.

And some tests.

### DIFF
--- a/src/dynamic_programming/fibonacci.rs
+++ b/src/dynamic_programming/fibonacci.rs
@@ -9,7 +9,7 @@ pub fn fibonacci(n: u32) -> u128 {
     // Use a and b to store the previous two values in the sequence
     let mut a = 0;
     let mut b = 1;
-    for i in 0..n {
+    for _ in 0..n {
         // As we iterate through, move b's value into a and the new computed
         // value into b.
         let c = a + b;
--- a/src/sorting/merge_sort.rs
+++ b/src/sorting/merge_sort.rs
@@ -129,4 +129,29 @@ mod tests {
         merge_sort(&mut v);
         assert_eq!(v, answer);
     }
+
+    #[test]
+    fn no_clone() {
+        #[derive(Debug, PartialEq, Eq)]
+        struct Point {
+            x: u32,
+            y: u32,
+        }
+
+        let mut v = vec![];
+
+        v.push(Point { x: 10, y: 9 });
+        v.push(Point { x: 3, y: 8 });
+        v.push(Point { x: 8, y: 7 });
+
+        merge_sort_cmp(&mut v, |l, r| l.x < r.x);
+        assert_eq!(v[0], Point { x: 3, y: 8 });
+        assert_eq!(v[1], Point { x: 8, y: 7 });
+        assert_eq!(v[2], Point { x: 10, y: 9 });
+
+        merge_sort_cmp(&mut v, |l, r| l.y < r.y);
+        assert_eq!(v[0], Point { x: 8, y: 7 });
+        assert_eq!(v[1], Point { x: 3, y: 8 });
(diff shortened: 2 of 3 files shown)


src/sorting/merge_sort.rs
```diff
@@ -0,0 +1,157 @@
...
+}
+
+/// Sort a slice using "merge sort" algorithm. `less` is a `FnMut(&T, &T) -> bool`, `less(&lhs,
+/// &rhs)` means called defined that `lhs` is less than `rhs`. The result order is undefined if
+/// this `less` function not define a total ordering for elements in the slice.
+///
+/// If `T` has zero sized, nothing happened.
+pub fn merge_sort_cmp<T, F>(v: &mut [T], mut less: F)
+where
+    F: FnMut(&T, &T) -> bool,
```
AnshulMalik: @plus7wist I love it, just one thing, do we really need the less function here?
We are requiring `T` to implement `Ord` trait, so can't we just use that?

plus7wist: Thanks, for you love that code.

- Do we need the `less` function?

  Maybe not. This is just meaningful if we make `merge_sort_cmp` as a part of the result of our merge sort implemention. This is good for a practical library, but not necessary for a repo built for educational purpose. So I feel it ok to remove this `pub`, make `merge_sort` the only pubilc function.

- We are requiring `T` to implement `Ord` trait.

  The good thing is `merge_sort` need that, but `merge_sort_cmp` don't. We can use `merge_sort_cmp` to sort slice of type `T: !Ord`. And, we can sort same type with different order. like this sample:

  ```rust
  struct Point {
    x: i32,
    y: i32,
  }

  let points = /* something */;
  merge_sort_cmp(&mut points, |lhs, rhs| lhs.x < rhs.x);
  merge_sort_cmp(&mut points, |lhs, rhs| lhs.y < rhs.y);
  ```

src/sorting/merge_sort.rs
```diff
@@ -0,0 +1,124 @@
...
+    sort_range(v, begin, middle, buf, less);
+    sort_range(v, middle, end, buf, less);
+    merge(v, begin, middle, end, buf, less);
+}
+
+fn merge<T, F>(v: &mut [T], begin: usize, middle: usize, end: usize, buf: *mut T, mut less: F)
+where
+    F: FnMut(&T, &T) -> bool,
+{
+    struct Range {
```
boydjohnson: I wouldn't declare the struct in the function. Please pull out the struct declaration and have it be module private.

src/sorting/merge_sort.rs
```diff
@@ -0,0 +1,124 @@
...
+        begin: begin,
+        end: middle,
+    };
+    let mut r = Range {
+        begin: middle,
+        end: end,
+    };
+    let mut top = unsafe { buf.add(begin) };
+
+    unsafe fn push<T>(top: &mut *mut T, data: *const T) {
```
boydjohnson: Same thing. Lets have `push` be a module private function.

src/dynamic_programming/fibonacci.rs
```diff
@@ -9,7 +9,7 @@ pub fn fibonacci(n: u32) -> u128 {
     // Use a and b to store the previous two values in the sequence
     let mut a = 0;
     let mut b = 1;
-    for i in 0..n {
+    for _ in 0..n {
```
boydjohnson: Let's take this out of the commit. We should address the compiler and clippy warnings in a separate PR.

plus7wist:
@boydjohnson Thanks for advice.

PR #86, opened 2019-09-30
implemented simple btree

lhiny:
Implemented a simple binary tree, under general/btree.rs


boydjohnson:
@lhiny Thank you for this PR. It looks like it is not a complete implementation. Also there was another PR that implemented a B-Tree (#102). 

You are welcome to find an unimplemented algorithm or data structure and implement that.

PR #92, opened 2019-10-14
Binary search:

crazcalm:
- Added check for is_empty to fix existing test
- Added a new test


crazcalm:
Thanks for creating/maintaining this project!

PR #104, opened 2019-11-11
Adding Rust merge_sort (with tests)

thepacketgeek:
Adding merge sort implementation, works for all Clone + Copy types. Also adding tests and updating the mod to include merge_sort.rs in test runs


thepacketgeek:
I'm going to move this to a branch

PR #113, opened 2019-12-17
Fix clippy warnings

boydjohnson:
With this PR there will be CI failures on clippy warnings. I think this gives greater guarantees of idiomatic code.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,5 +5,5 @@ before_script:
   - rustup component add clippy
 script:
   - cargo fmt --all -- --check
-  - cargo clippy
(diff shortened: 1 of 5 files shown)


boydjohnson:
Hi @AnshulMalik, There were a small amount of clippy lint that `clippy --all` complains about that are fixed in this PR. There is a fix for an incorrect impl of `is_empty()` on Heap. The main point of this PR is to turn on `cargo clippy --all -- -D warnings` in CI.

AnshulMalik (approved):

AnshulMalik:
Thanks @boydjohnson

PR #115, opened 2019-12-19
Add longest increasing subsequence algorithm

pgimalac:


boydjohnson:
Hi @pgimalac, When you deal with the merge conflict can you update your git commit message to "Add longest increasing subsequence algorithm".

siriak (approved):
Good, but need to fix linting errors

hedonihilist:
Hi guys, I've read the code and I think it's correct in terms of calculating the length of the longest increasing subsequence, but can produce wrong sequence. Here is a test case:
```
    fn test() {
        let data = vec![1,3,5,7,9,4];
        assert_eq!(vec![1,3,5,7,9], longest_increasing_subsequence(&data));
    }
```
which will fail with error:
```
thread 'tests::test' panicked at 'assertion failed: `(left == right)`
  left: `[1, 3, 5, 7, 9]`,
 right: `[1, 3, 4, 7, 9]`', src/lib.rs:64:9
```

boydjohnson:
While the algorithm is close to correct, I don't think it is efficient for the problem. The [wikipedia page for longest increasing subsequence](https://en.wikipedia.org/wiki/Longest_increasing_subsequence) describes an algorithm that maintains two arrays of indexes, that it describes as efficient.

pgimalac:
@hedonihilist indeed it seems there is an error somewhere, this is embarrassing !

I don't remember where I found the pseudocode to write this Rust version ; I'll take a look at the wikipedia algorithm and possibly rewrite everything.

pgimalac:
I realized while trying to merge with master that another pull request for the same algorithm has already been accepted, so I guess I can close this one

imp2002:
oh i‘m sorry to that😱 the same algorithm has already been accepted should be `the Longest continuous increasing subsequence` not `the Longest increasing subsequence`  it's my mistake. After the pr (#216 ) accepted, you can merge with master and continue to this great work.

imp2002 (approved):

siriak:
What's the matter with the workflow files? They shouldn't be changed afaik

imp2002 (changes requested):
the committ log

pgimalac:
Yikes indeed, I think I know what I did wrong (although I'm not completely sure), I'll try to clean this up 😅

siriak (approved):
LGTM, thanks!

pgimalac:
So... this time it really has already been done (and merged) by someone else right ? (https://github.com/TheAlgorithms/Rust/pull/235)

PR #116, opened 2019-12-19
convex hull algorithm: Graham scan

pgimalac:


src/general/convex_hull.rs
```diff
@@ -0,0 +1,172 @@
...
+                // distance (we want the closest to be first)
+                *x,
+            )
+        })
+        .collect();
+    points.sort_by(|a, b| a.partial_cmp(b).unwrap_or(Equal));
+    return points.into_iter().map(|x| x.2).collect();
+}
+
+fn produit_vectoriel(a: &(f64, f64), b: &(f64, f64), c: &(f64, f64)) -> f64 {
```
boydjohnson: Hi @pgimalac This is calculating the z-coordinate of the vector product, if I am not mistaken. So a function name that incorporates that the z-coordinate is what is being returned would be best. Such as `calc_z_coord_vector_product`.

I notice that the function name is in French. I think it would be best for consistency that the function name should be in English, since the other text is in English.

src/general/convex_hull.rs
```diff
@@ -0,0 +1,172 @@
...
+    for point in points {
+        while stack.len() > 1
+            && produit_vectoriel(&stack[stack.len() - 2], &stack[stack.len() - 1], &point) < 0.
+        {
+            stack.pop();
+        }
+        stack.push(point);
+    }
+
+    return stack;
```
boydjohnson: The return is not necessary. Clippy calls it a `needless return`. In the future it will be a hard ci failure from clippy.

src/general/convex_hull.rs
```diff
@@ -0,0 +1,172 @@
...
+                ((x.1 - min.1) as f64).atan2((x.0 - min.0) as f64),
+                // angle
+                ((x.1 - min.1) as f64).hypot((x.0 - min.0) as f64),
+                // distance (we want the closest to be first)
+                *x,
+            )
+        })
+        .collect();
+    points.sort_by(|a, b| a.partial_cmp(b).unwrap_or(Equal));
+    return points.into_iter().map(|x| x.2).collect();
```
boydjohnson: This return is not necessary. Clippy calls it a `needless return`. In the future it will be a hard ci error.

src/general/convex_hull.rs
pgimalac: Indeed I was reading the french wikipedia article while writing the algorithm and I didn't even notice I wrote that in french ! And that's exactly what's calculated yes.

PR #134, opened 2020-06-28
Add Graph container, depth and breadth traversal

prateeknischal:
Add a module to create a graph. Implement Depth and Breadth first
traversal algorithms.


DaGuich:
This PR is handling the graphs data structure and BFS mentioned in #3

PR #136, opened 2020-07-08
Add genetic algorithm and a xorshift implementation to avoid rand crate

D4rkia:
Just added a genetic algorithm folder with a basic algorithm inside for string evolution and a xorshift algorithm to avoid need for the external rand crate

- [X]  cargo fmt
- [X]  cargo test


src/general/xorshift.rs
```diff
@@ -0,0 +1,95 @@
...
+    }
+
+    #[test]
+    fn rand_float_distribution() {
+        // Simple number distribution calculation using 0.0 to 1.0 values
+        // to verify random distribution of the output
+        let mut rng = Rand::new(0);
+        let container: Vec<f32> = (0..500_000).map(|_| rng.rand_float_normal()).collect();
+        assert_eq!(
+            (container.iter().filter(|&n| *n > 0.5).count() as f32 / 500_000.0) < 5.5,
```
AnshulMalik, line 87: this condition will always be `<5.5` isn't it? do we want to do `<0.5` ?

D4rkia: You're right i've fixed this and the next one.

PR #141, opened 2020-07-28
[Feat] Add Radix Sort

rytheo:
A radix sort implementation for issue #3.

I considered using the existing `counting_sort`, but I can't tell it what radix and digit to use when sorting.

### DIFF
--- a/src/sorting/radix_sort.rs
+++ b/src/sorting/radix_sort.rs
@@ -11,13 +11,7 @@ pub fn radix_sort(arr: &mut [u64]) {
         None => return,
     };
     // Make radix a power of 2 close to arr.len() for optimal runtime
-    let radix = {
-        let mut r = 2;
-        while r < arr.len() {
-            r *= 2;
-        }
-        r
-    };
+    let radix = arr.len().next_power_of_two();
     // Counting sort by each digit from least to most significant
     let mut place = 1;
     while place <= max {
--- a/src/sorting/radix_sort.rs
+++ b/src/sorting/radix_sort.rs
@@ -1,0 +1,68 @@
+/// Sorts the elements of `arr` in-place using radix sort.
+///
+/// Time complexity is `O((n + b) * logb(k))`, where `n` is the number of elements,
+/// `b` is the base (the radix), and `k` is the largest element.
+/// When `n` and `b` are roughly the same maginitude, this algorithm runs in linear time.
+///
+/// Space complexity is `O(n + b)`.
+pub fn radix_sort(arr: &mut [u64]) {
+    let max: usize = match arr.iter().max() {
+        Some(&x) => x as usize,
+        None => return,
+    };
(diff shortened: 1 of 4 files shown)


src/sorting/radix_sort.rs
```diff
@@ -0,0 +1,68 @@
...
+/// When `n` and `b` are roughly the same maginitude, this algorithm runs in linear time.
+///
+/// Space complexity is `O(n + b)`.
+pub fn radix_sort(arr: &mut [u64]) {
+    let max: usize = match arr.iter().max() {
+        Some(&x) => x as usize,
+        None => return,
+    };
+    // Make radix a power of 2 close to arr.len() for optimal runtime
+    let radix = {
```
xfix, line 14:
```suggestion
    let radix = arr.len().next_power_of_two();
```

src/sorting/radix_sort.rs
```diff
@@ -0,0 +1,62 @@
...
+/// When `n` and `b` are roughly the same maginitude, this algorithm runs in linear time.
+///
+/// Space complexity is `O(n + b)`.
+pub fn radix_sort(arr: &mut [u64]) {
+    let max: usize = match arr.iter().max() {
+        Some(&x) => x as usize,
+        None => return,
+    };
+    // Make radix a power of 2 close to arr.len() for optimal runtime
+    let radix = arr.len().next_power_of_two();
```
AnshulMalik, line 14: How does this help for optimal runtime?

src/sorting/radix_sort.rs
```diff
@@ -0,0 +1,62 @@
...
+    while place <= max {
+        let digit_of = |x| x as usize / place % radix;
+        // Count digit occurrences
+        let mut counter = vec![0; radix];
+        for &x in arr.iter() {
+            counter[digit_of(x)] += 1;
+        }
+        // Compute last index of each digit
+        for i in 1..radix {
+            counter[i] += counter[i - 1];
```
AnshulMalik, line 26: how does this for loop adds into the complexity?

src/sorting/radix_sort.rs
rytheo: It's a simple way to choose a radix of the same magnitude as `n`, which is necessary for the algorithm to run in linear time.

src/sorting/radix_sort.rs
rytheo: This loop has `O(b)` runtime where `b` is the radix. Therefore, this loop contributes the `b` of the `(n + b)` factor in the overall complexity.

AnshulMalik: I see, this loop iterates `radix` times, not base of radix times, or am i missing something?

AnshulMalik: okay got it.

AnshulMalik (approved):

PR #145, opened 2020-10-03
Add project euler directory and problem 3 solution

MonliH:
This PR adds a `project_euler` directory and a few guidelines to the respective readme. Project Euler solutions were originally requested in #3.

 It also adds a solution for [problem 3](https://projecteuler.net/problem=3).

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -5,6 +5,7 @@ pub mod general;
 pub mod searching;
 pub mod sorting;
 pub mod string;
+pub mod project_euler;
 
 #[cfg(test)]
 mod tests {
--- a/src/project_euler/README.md
+++ b/src/project_euler/README.md
@@ -1,0 +1,4 @@
+# Project Euler
+
+Put any solutions to problems from [project euler](https://projecteuler.net/)
+here.
--- a/src/project_euler/README.md
+++ b/src/project_euler/README.md
@@ -2,3 +2,16 @@
 
 Put any solutions to problems from [project euler](https://projecteuler.net/)
 here.
+
+# Guidelines
+
+Please provide a link and the number of the problem in a module level doc 
+comment. For example:
+
+```rust
+//! Larget Prime Factor
+//! 
+//! Problem #3.
+//! https://projecteuler.net/problem=3
+```
+
--- a/src/project_euler/largest_prime_factor.rs
+++ b/src/project_euler/largest_prime_factor.rs
@@ -1,6 +1,10 @@
+//! Larget Prime Factor
+//! 
+//! Problem #3.
+//! https://projecteuler.net/problem=3
 
 /// Get the largest prime factor of the (unsigned) integer `n`
-fn largest_prime_factor(mut n: u64) -> u64 {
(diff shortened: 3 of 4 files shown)


AnshulMalik:
Thanks @MonliH .

Just a few comments: 
Since there can be multiple solutions for a problem in project euler, we should have a directory for each problem and place solutions inside them.
It will also be easy for people to find the programs for a particular problem if they are numbered.

MonliH:
>  It will also be easy for people to find the programs for a particular problem if they are numbered.

Yes, I was thinking about numbering them. `_3_largest_prime_factor.rs`  (you can't import modules starting with numbers) doesn't look too appealing to me. Putting the number (e.g. `largest_prime_factor_3.rs`) at the end looks cleaner but the problem is you don't get the benefit of sorting by filename, though. I suggest we go with the second approach, what do you think?

>  Since there can be multiple solutions for a problem in project euler, we should have a directory for each problem and place solutions inside them.

Hmm, seems to me this would be unnecessary in most cases, cause then we'd end up with a bunch of directories with just a `mod.rs` file. Maybe if someone wants to contribute another solution they can turn the file-based module into a folder-based one? (we can add instructions to the readme)

AnshulMalik:
We should be consistent at things, keeping two ways will most likely confuse people. 
I was thinking about naming directory as numbers, so it'll be easy to find solutions, having just `mod.rs` shouldn't be an issue if people can find what they are looking for easily.

What do you think?

MonliH:
Sure, that sounds good, actually. I'll document it on the readme and move to that format.

growingspaghetti:
Let me in! https://github.com/growingspaghetti/project-euler

MonliH:
@AnshulMalik I just realized, you can't have folders start with numbers and use them with rust. It would have to be `_<number>` or similar. Maybe `<short_problem_desc>_<number>`?

AnshulMalik:
Sure, we can do `_number`

AnshulMalik:
@growingspaghetti, yay, please feel free to raise pull requests for the solutions after this structure is finalized :)

MonliH:
> Sure, we can do `_number`

Awesome, changing the structure right now.

src/project_euler/_3/mod.rs
```diff
@@ -0,0 +1,36 @@
...
+//! https://projecteuler.net/problem=3
+
+/// Get the largest prime factor of the (unsigned) integer `n`
+pub fn largest_prime_factor(mut n: u64) -> u64 {
+    // Largest prime factor
+    let mut lpf = 2;
+    while n > lpf {
+        if n % lpf == 0 {
+            n /= lpf;
+            lpf = 2;
```
AnshulMalik, line 13: why do we have `lpf=2` here?

MonliH: Because if we don't reset it, the loop would never terminate (try it!). Notice the while loop condition. The way this algorithm works is by dividing `n` by primes each time (`lpf` is always prime, unless it's `1`, because `2` is a prime, and if it can ever divide perfectly into `n`), untill `n` is equal to or less than `lpf`.

PR #149, opened 2020-10-06
space efficient edit distance algorithm

TianyiShi2001:
Added a space-efficient edit distance algorithm.

The original one uses **O(nm)** space. This one uses **O(n)** space and is slightly faster.

### DIFF
--- a/src/dynamic_programming/edit_distance.rs
+++ b/src/dynamic_programming/edit_distance.rs
@@ -40,6 +40,10 @@ pub fn edit_distance(str_a: &str, str_b: &str) -> u32 {
 
 /// The space efficient version of the above algorithm.
 ///
+/// Instead of storing the `m * n` matrix expicitly, only one row (of length `n`) is stored.
+/// It keeps overwriting itself based on its previous values with the help of two scalars,
+/// gradually reaching the last row. Then, the score is `matrix[n]`.
+///
 /// # Complexity
 ///
 /// - time complexity: O(nm),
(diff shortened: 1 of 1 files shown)


src/dynamic_programming/edit_distance.rs
```diff
@@ -31,27 +38,76 @@ pub fn edit_distance(str_a: &str, str_b: &str) -> u32 {
...
+    let (str_a, str_b) = (str_a.as_bytes(), str_b.as_bytes());
+    let (m, n) = (str_a.len(), str_b.len());
+    let mut dp_matrix: Vec<u32> = vec![0; n + 1]; // the dynamic programming matrix (only 1 row stored)
+    let mut s: u32; // dp_matrix[i - 1][j - 1] or dp_matrix[i - 1][j]
+    let mut c: u32; // dp_matrix[i][j - 1] or dp_matrix[i][j]
+    let mut a: u8; // str_a[i - 1]
+    let mut b: u8; // str_b[j - 1]
+
+    // 0th row
+    for j in 1..=n {
```
AnshulMalik, line 63: Looks good, it'll be easier to understand if the variable names depict their purpose.

TianyiShi2001: I think using `m` and `n` to represent the number of columns and rows of a matrix and using `M[i][j]` to refer to a cell in a matrix is quite conventional. The only variable names that might be confusing are `s` and `c`. It is difficult to give descriptive names for them because they mean different positions in the matrix depending on the context. I've explained them in more detail in https://github.com/TheAlgorithms/Rust/pull/149/commits/058507ed0bf745cdc408d0cd46516445fc5c8b83

AnshulMalik (approved):

PR #158, opened 2020-10-14
Make unsafe blocks smaller

rhysd:
As the book says at https://doc.rust-lang.org/book/ch19-01-unsafe-rust.html,

> People are fallible, and mistakes will happen, but by requiring these five unsafe operations to be inside blocks annotated with unsafe you’ll know that any errors related to memory safety must be within an unsafe block. Keep unsafe blocks small; you’ll be thankful later when you investigate memory bugs.

`unsafe` block should be as small as possible. Since this repository is a good showcase for learning Rust, I think following best practices is important. This PR changes usage of `unsafe` block to make them smaller.

### DIFF
--- a/src/data_structures/linked_list.rs
+++ b/src/data_structures/linked_list.rs
@@ -40,20 +40,17 @@ impl<T> LinkedList<T> {
 
     pub fn add(&mut self, obj: T) {
         let mut node = Box::new(Node::new(obj));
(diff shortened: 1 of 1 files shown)


AnshulMalik (approved):

AnshulMalik:
Thanks @rhysd.

PR #159, opened 2020-10-14
Add another implementation of rot13

iakgoog:


imp2002 (approved):
No problem with the code, thanks. But it implements like `caesar.rs`, i don't thank it's another and the name `another_rot13` may not be very good.

siriak (approved):
Rot13 is a special case of Caesar cipher, that's why they look similar :) LGTM

PR #165, opened 2020-10-23
Add Rabin Karp and Finite Automata

MovAh13h:
Added Rabin Karp pattern Matching algorithm

### DIFF
--- a/src/string/README.md
+++ b/src/string/README.md
@@ -9,9 +9,17 @@ __Properties__
 * Case space complexity  O(w)
 
 ### [Rabin Karp](./rabin_karp.rs)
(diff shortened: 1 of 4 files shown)


GandalfTheGrayOfHell:
The other PR for the same fails tests and makes a mistake. In Rust `std::string::String` is UTF8 compatible which was not accounted for by setting a variable incorrectly. This PR takes care of both the things.

@AnshulMalik

horahh (approved):
Great input overall, would be nice to handle error instead of unwrap with the ? operator

GandalfTheGrayOfHell:
Most of the `unwrap` calls are actually for an `Option` and not a `Result`. I dont think that should be a problem.

PR #174, opened 2020-12-08
Move tests for bubble sort to match standard

horahh:
* Split test in two ascending and descending

### DIFF
--- a/.gitconfig
+++ b/.gitconfig
@@ -1,0 +1,2 @@
+[core]
+    hooksPath = git_hooks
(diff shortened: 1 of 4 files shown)


horahh:
can someone point me to where to look for the failure? I am not seeing any information about what the regression is? and in my machine cargo test is passing the 100 tests

sandmor:
> The command "cargo fmt --all -- --check" exited with 1.

The PR fails by the format not by the test, to fix it just run `cargo fmt --all`

AnshulMalik (approved):
LGTM

AnshulMalik:
Thanks @horahh

PR #176, opened 2021-01-11
Added 0-1 Knapsack

endygamedev:
Add 0-1 Knapsack algorithm and tests


src/dynamic_programming/knapsack.rs
```diff
@@ -0,0 +1,157 @@
...
+///     * `weights` - set of weights for each item
+///     * `values` - set of values for each item
+///
+/// Complexity
+///     - time complexity: O(nw),
+///     - space complexity: O(nw),
+///
+/// where `n` and `w` are `number of items` and `knapsack capacity`
+pub fn knapsack(
+    n: usize,
```
AnshulMalik, line 71: Can we not deduce n from the size of `weights` and `values`?

src/dynamic_programming/knapsack.rs
```diff
@@ -0,0 +1,157 @@
...
+/// Arguments:
+///     * `weights` - set of weights for each item
+///     * `m` - knapsack table with maximum values
+///     * `i` - include items 1 through `i` in knapsack (for the initial value, use `n`)
+///     * `j` - maximum weight of the knapsack
+fn knapsack_items(weights: &[usize], m: &[Vec<usize>], i: &usize, j: &usize) -> Vec<usize> {
+    if *i == 0 {
+        return vec![];
+    }
+    if m[*i][*j] > m[*i - 1][*j] {
```
AnshulMalik, line 47: I see a lot of dereferencing in this function, can this be avoided?

src/dynamic_programming/knapsack.rs
endygamedev: I think it can be done. I was just trying to make functions close to their mathematical formulation, but we can take the length of the `weights` and `values`.
I can redo this.

src/dynamic_programming/knapsack.rs
endygamedev: Okay, I'll see what I can do about it.
Thank you!

AnshulMalik:
@endygamedev It'd be wise if we update the readme as well, that this is implemented

AnshulMalik (approved):

PR #179, opened 2021-02-01
Add Levenshtein distance algorithm to string algorithms

AltriusRS:
This is my first contribution, but i felt like the algorithm should be included, so i implemented it and listed it under the string methods.


Volpe95:
It's failing the checks , also you have to edit the Readme file to include your newly added algorithm.

AltriusRS:
i will happily edit the readme, the checks failing do not appear to be an issue with my code, it compiles and runs just fine, and passes all of the clippy checks just fine. Maybe the probem with the checks is that you are running Javascript checks on Rust code?

PR #184, opened 2021-03-03
This stack implementation supports push, pop, and peek methods.

dash2507:


dash2507:
Hi @AnshulMalik, sorry to bug you. But I didn't know anything else to get this PR merged so can you check this PR?. Thank you.

src/data_structures/stack.rs
```diff
@@ -0,0 +1,158 @@
...
+}
+
+impl<T> Stack<T> {
+    pub fn new() -> Self {
+        Self { head: None, top: 0 }
+    }
+
+    /// Push an item to the top of a stack.
+    pub fn push(&mut self, value: T) {
+        let mut node = Box::new(Node::new(value));
```
pickfire, line 29: This will be in the heap rather than stack.

src/data_structures/stack.rs
```diff
@@ -0,0 +1,158 @@
+use std::fmt::{self, Display, Formatter};
+use std::ptr;
+use std::ptr::NonNull;
+
+struct Node<T> {
+    val: T,
+    next: Option<NonNull<Node<T>>>,
```
pickfire, line 7: Why not just use `Option<Box<Node<T>>>`?

src/data_structures/stack.rs
dash2507: what are you suggesting then?  because, as I know there should be a push method on stack right?

src/data_structures/stack.rs
dash2507: very good point. I will change that.

src/data_structures/stack.rs
pickfire: I mean we are working with stack, but we are using the heap, or it is just the stack data stucture but stuff works in heap?

dash2507: sorry for the confusion, it is a stack data structure which stores data on the heap.

PR #185, opened 2021-03-18
[fibonacci] fix comment

p00f:
Overflow occurs at n = 188, fib(187) is returned correctly
image


imp2002 (changes requested):
I texted the `185` and `186`, it's really overflow the 128-bit unsigned integer at n=186, not `188`.

At the begin of the code, the athour says this function uses the definition of Fibonacci where:
`F(0) = F(1) = 1 and F(n+1) = F(n) + F(n-1) for n>0`

maybe you use the other definition, so get the different reault.

imp2002:
@siriak maybe this PR can be close😊

PR #193, opened 2021-05-13
Add Queue

HakkerBarry:
Simple implementation of the queue using vec


boydjohnson:
@HakkerBarry I think the CI issue is that the struct Queue isn't `pub`. If you make it `pub struct Queue` the CI may pass.

PR #196, opened 2021-05-17
fix typo and add shell sort to src/sorting/mod.rs

keob:
1. fix shell sort typo
2. add shell sort to src/sorting/mod.rs


imp2002 (approved):
LGTM! Thanks.

PR #200, opened 2021-06-07
Refactor BTreeProps::new, reducing verbosity.

Seideun:
I found extraneous uses of matches and if-else's in the code, many of which can be replaced by chain-call styles. I think that improves readability and is more Rustacean. XD


siriak (approved):
LGTM, thanks!

siriak:
Could you run `cargo fmt --all -- --check` to see and fix build errors?

Issue #188, opened 2021-04-05
Rewrite `Heap.rs`


k12ish:
This trivial test fails:
```rust
#[test]
fn test_empty_heap() {
    let mut heap = MaxHeap::new::<i32>();
    assert_eq!(heap.next(), None)
}
```

    ---- data_structures::heap::tests::test_empty_heap stdout ----
    thread 'data_structures::heap::tests::test_empty_heap' panicked at 'attempt to subtract with
    overflow', src/data_structures/heap.rs:117:9

Other big issues:

- `Vec` contains a length meaning that `count` is unneeded
-  The trait bound `Default` can be entirely removed with zero-based indexing, saving memory and making the implementation more flexible

I am more than happy to enact these changes, if that is acceptable to you.

siriak:
Hi, you can make a PR to fix that and I'll get it reviewed and merged. Thanks for sharing!

plaxi0s:
Hi I made to [PR](https://github.com/TheAlgorithms/Rust/pull/215) to fix this. Let me know if any other changes are required. 
Thanks!

siriak:
Cool, thanks! Feel free to make more PRs if you want to fix something else

PR #215, opened 2021-08-31
Fixes  TheAlgorithms/Rust#188

plaxi0s:


imp2002 (approved):
Yeah, `if count == 0` then `count -= 1` cause the usize type number become a negative number.

Your change is right， thanks！

PR #234, opened 2021-10-09
Add solution for Problem 1 of Project Euler

bernhardfritz:


siriak:
Hi, we don't include project Euler solutions, thank you!

siriak:
Good point, I'll remove it

PR #246, opened 2021-10-24
add gnome_sort with tests

donbarbos:
* add gnome sort
* add a description of the sorting algorithm in README.md


DaveAxiom:
I'm the original author for the NQueens solution and I saw this patch to it. It appears when I wrote the solution, I had debugging information printed and I followed some best practice advice on how to create a `do .. while` loop in Rust. I removed the debug printing and it created a leftover structural logic issue I missed. It appears that only now the Rust checker invalidated a code implementation that was once recommended as a best practice! I guess I must use `loop{}` in the future. I'm a programming language polyglot having written in 20+ programming languages and I think that Rust is making a mistake in not including a `do .. while` loop.

PR #251, opened 2021-10-31
added some simple math algorithms

DONSIMON92:
* added implementation of prime and perfect numbers and prime factorization with tests
* added new files to ` DIRECTORY.md`
* updated `CONTRIBUTING.md` because there was no clippy linter info


DIRECTORY.md
```diff
@@ -1,4 +1,4 @@
-# List of all files
+#numberi List of all files
```
siriak, line 1: Why change?

---
**Sema Reaction:** :question: I have a question

siriak (changes requested):

DIRECTORY.md
DONSIMON92: thanks for the correction
i changed my last commit

siriak (approved):

PR #257, opened 2021-11-12
Bucket sort implementation + inplace insertion sort

tuqqu:
This is a `bucket sort` algorithm implementation. 
As it is best to use `insertion sort` as a  subroutine of `bucket sort`, I went forward and improved the `insertion sort` version to one that sorts in-place.

So this PR includes:
* `bucket sort`
* in-place `insertion sort` re-write


siriak (approved):
Great job, thanks!

PR #258, opened 2021-11-15
Add stack using singly linked list

Aisuko:
Signed-off-by: aisuko <urakiny@gmail.com>

Add stack using singly linked list and test case.

### DIFF
--- a/src/data_structures/stack_using_singly_linked_list.rs
+++ b/src/data_structures/stack_using_singly_linked_list.rs
@@ -1,0 +1,254 @@
+// the public struct can hide the implementation detail
+pub struct Stack<T> {
+    head: Link<T>,
+}
+
+type Link<T> = Option<Box<Node<T>>>;
+
+struct Node<T> {
+    elem: T,
+    next: Link<T>,
+}
+
+impl<T> Stack<T> {
+    // Self is an alias for Stack
+    // We implement associated function name new for single-linked-list
+    pub fn new() -> Self {
+        // for new function we need to return a new instance
+        Self {
+            // we refer to variants of an enum using :: the namespacing operator
+            head: None,
+        } // we need to return the variant, so there without the ;
+    }
+
+    // As we know the primary forms that self can take: self, &mut self and &self, push will change the linked list
+    // so we need &mut
+    // The push method which the signature's first parameter is self
+    pub fn push(&mut self, elem: T) {
+        let new_node = Box::new(Node {
+            elem,
+            next: self.head.take(),
+        });
+        // don't forget replace the head with new node for stack
+        self.head = Some(new_node);
+    }
+    ///
+    /// In pop function, we trying to:
+    /// * check if the list is empty, so we use enum Option<T>, it can either be Some(T) or None
+    ///   * if it's empty, return None
+    ///   * if it's not empty
+    ///     * remove the head of the list
+    ///     * remove its elem
+    ///     * replace the list's head with its next
+    ///     * return Some(elem), as the situation if need
+    ///
+    /// so, we need to remove the head, and return the value of the head
+    pub fn pop(&mut self) -> Result<T, &str> {
+        match self.head.take() {
+            None => Err("Stack is empty"),
+            Some(node) => {
+                self.head = node.next;
+                Ok(node.elem)
+            }
+        }
+    }
+
+    pub fn is_empty(&self) -> bool {
+        // Returns true if the option is a [None] value.
+        self.head.is_none()
+    }
+
+    pub fn peek(&self) -> Option<&T> {
+        // Converts from &Option<T> to Option<&T>.
+        match self.head.as_ref() {
+            None => None,
+            Some(node) => Some(&node.elem),
+        }
+    }
+
+    pub fn peek_mut(&mut self) -> Option<&mut T> {
+        match self.head.as_mut() {
+            None => None,
+            Some(node) => Some(&mut node.elem),
+        }
+    }
+
+    pub fn into_iter_for_stack(self) -> IntoIter<T> {
+        IntoIter(self)
+    }
+    pub fn iter(&self) -> Iter<'_, T> {
+        Iter {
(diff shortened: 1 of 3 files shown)


src/data_structures/single_linked_list.rs
```diff
@@ -0,0 +1,248 @@
...
+        } // we need to return the variant, so there without the ;
+    }
+
+    // As we know the primary forms that self can take: self, &mut self and &self, push will change the linked list
+    // so we need &mut
+    // The push method which the signature's first parameter is self
+    pub fn push(&mut self, elem: T) {
+        let new_node = Box::new(Node {
+            elem,
+            // mem::replace function lets us steal a value out of a borrow by replacing it with another value
```
tuqqu, line 30: I don't think `mem::replace` is used anywhere here (probably an old comment)

src/data_structures/single_linked_list.rs
```diff
@@ -0,0 +1,248 @@
...
+        }
+    }
+}
+
+/// Rust has nothing like a yield statement, and there's actually 3 different kinds of iterator should to implement
+
+// Collections are iterated in Rust using the Iterator trait, we define a struct implement Iterator
+pub struct IntoIter<T>(SingleLinkedList<T>);
+
+impl<T> SingleLinkedList<T> {
```
tuqqu, line 100: (here and below) is it really a good idea to use separate `impls X` in the same module?

src/data_structures/single_linked_list.rs
```diff
@@ -0,0 +1,248 @@
...
+    /// * check if the list is empty, so we use enum Option<T>, it can either be Some(T) or None
+    ///   * if it's empty, return None
+    ///   * if it's not empty
+    ///     * remove the head of the list
+    ///     * remove its elem
+    ///     * replace the list's head with its next
+    ///     * return Some(elem), as the situation if need
+    ///
+    /// so, we need to remove the head, and return the value of the head
+    pub fn pop(&mut self) -> Option<T> {
```
tuqqu, line 47: Wouldn't it be better to return `Result<T, UnderflowError>` and error when the list/stack is empty?

src/data_structures/single_linked_list.rs
```diff
@@ -0,0 +1,248 @@
...
+}
+
+type Link<T> = Option<Box<Node<T>>>;
+
+struct Node<T> {
+    elem: T,
+    next: Link<T>,
+}
+
+impl<T> SingleLinkedList<T> {
```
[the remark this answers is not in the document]
Aisuko: Thanks for reviewing and comment details. You're right. I will replace the name to `Stack using Singly Linked List`.

src/data_structures/single_linked_list.rs
Aisuko: Yes, this should be remove.

src/data_structures/single_linked_list.rs
Aisuko: Here, I will group them into one basic `impl` above.

src/data_structures/single_linked_list.rs
```diff
@@ -0,0 +1,248 @@
...
+    // Self is an alias for SingleLinkedList
+    // We implement associated function name new for single-linked-list
+    pub fn new() -> Self {
+        // for new function we need to return a new instance
+        SingleLinkedList {
```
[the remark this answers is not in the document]
Aisuko: I will replace `SingleLinkedList` to `Self` to stay consistent, thank you.

src/data_structures/single_linked_list.rs
Aisuko: Agree with you, but here `pop` return `Option<T>` to support we implement `next(&mut self) -> Option<Self::Item>` method `trait Iterator` at below. And we can do this on `peek` function.

So, please help take a review. thanks.
```
    pub fn peek(&self) -> Result<&T, &str> {
        // Converts from &Option<T> to Option<&T>.
        match self.head.as_ref() {
            None => Err("Stack is empty"),
            Some(node) => Ok(&node.elem),
        }
    }
```

tuqqu: * `pop()` [should](https://en.wikipedia.org/wiki/Stack_(abstract_data_type)#Linked_list) really error on empty stacks. If you want to use it in iteration, you can deal with error handling there, converting `UnderflowError` to `None`.
* `peek()` may or may not error depending on implementation. But it's better to be consistent and do whatever is chosen with `peek_mut()` as well.

src/data_structures/stack_using_singly_linked_list.rs
```diff
@@ -0,0 +1,254 @@
...
+
+/// The drop method of single-linked-list. There's a question that do we need to worry about cleaning up our list?
+/// As we all know the ownership and borrow mechanism, so we know the type will clean automatically after it goes out the scope,
+/// this implement by the Rust compiler automatically did which mean add trait `drop` for the automatically.
+///
+/// So, the complier will implements Drop for `List->Link->Box<Node> ->Node` automatically and tail recursive to clean the elements
+/// one by one. And we know the recursive will stop at Box<Node>
+/// https://rust-unofficial.github.io/too-many-lists/first-drop.html
+///
+/// As we know we can't drop the contents of the Box after deallocating, so we need to manually write the iterative drop
```
siriak, line 106: Please fix spelling mistakes

siriak (changes requested):

src/data_structures/stack_using_singly_linked_list.rs
Aisuko: Done

src/data_structures/single_linked_list.rs
Aisuko: No problem, the `pop()` using `Result<T, Err>` and return `Err` while the list empty. And the `peek()` we follow the style like `peek_mut()`

siriak (approved):
LGTM, thanks!

PR #261, opened 2021-12-19
Add Transposition Cipher

NashJames:
Hey! I decided to have a go at this.

I opted for a slightly more complex implementation of it, that would allow multiple keywords. I figured it'd be a shame not to incorporate it given the fairly big historical impact of the double columnar. Plus, it has an added bonus of being slightly more robust.

Give me a shout if anything is trouble.


imp2002 (approved):
LGTM! Thanks.🎈

PR #268, opened 2022-01-21
Fix queue implementation

### LINKED ISSUE #266: Queue is not correctly implemented
Hello, maybe im wrong, (im new to rust) but seeing the queue code, the dequeue operation
is taking O(n) time, and its suposed to take O(1) time.

I assume that the vector is implemented like in other languages (e.g c++) and a vector is really a dynamic array, and remove the first take O(n) because you need to swap all the elements one position to the left.

If this assumption is false, then ignore this comment.

rytheo:
Use a LinkedList instead of a Vec for enqueue/dequeue operations in O(1) time.

Closes #266

### DIFF
--- a/src/data_structures/queue.rs
+++ b/src/data_structures/queue.rs
@@ -1,35 +1,30 @@
+use std::collections::LinkedList;
+
 #[derive(Debug)]
 pub struct Queue<T> {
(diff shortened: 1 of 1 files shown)


siriak (approved):
LGTM, thanks!

PR #269, opened 2022-01-21
Fix leaky linked list

### LINKED ISSUE #242: LinkedList leaks Node memory when dropped
The [linked_list](https://github.com/TheAlgorithms/Rust/blob/master/src/data_structures/linked_list.rs) module stores references to nodes using `NonNull`s that are never dropped at any point.
Unless there is something I missed, this code doesn't ever free the memory allocated to the nodes.

rytheo:
Added a `pop_front` method that reconstructs boxed nodes so they can be dropped, then used that method to implement `Drop` so that all nodes are properly deallocated.

Closes #242

### DIFF
--- a/src/data_structures/linked_list.rs
+++ b/src/data_structures/linked_list.rs
@@ -1,4 +1,5 @@
 use std::fmt::{self, Display, Formatter};
+use std::marker::PhantomData;
 use std::ptr::NonNull;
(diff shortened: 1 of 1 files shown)


imp2002 (approved):
LGTM! Thansk.🎈

PR #273, opened 2022-02-17
Feature/add tim sort

hosseind88:


hosseind88:
Hi Guys, sorry can you please help me to understand why my code doesn't work in a specific situation?
here as j is usize and can't be negative, I added an if statement to break from the loop when j goes to 0
it does sorting well but it ignores 1 loop iteration and the sorted array would be wrong

Screen Shot 2022-02-17 at 12 55 44
Screen Shot 2022-02-17 at 10 59 45

hosseind88:
okay I fixed it but it's a little dirty I guess, please inform me about the improvements I can add to make it better😅, thanks

siriak:
The `if` is required there because `j` could underflow if `left` is `0`

siriak (changes requested):
Please see and fix errors from clippy

hosseind88:
thanks, fixed.
so you think that the code is ok?

PR #275, opened 2022-02-22
performance_improvement of math::prime_check

fffzlfk:
Improved performance of math::prime_check using Optimized School Method.

### DIFF
--- a/src/math/prime_check.rs
+++ b/src/math/prime_check.rs
@@ -1,13 +1,19 @@
 pub fn prime_check(num: usize) -> bool {
-    if (num > 1) & (num < 4) {
+    if num <= 1 {
+        return false;
+    }
+    if num <= 3 {
         return true;
-    } else if (num < 2) || (num % 2 == 0) {
+    }
+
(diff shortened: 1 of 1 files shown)


imp2002:
This can run more fast indeed, but it's not look like a radical optimization, more like a little trick. Anyway thanks your comtribute.

imp2002 (changes requested):
This can run more fast indeed, but it's not look like a radical optimization, more like a little trick. Anyway thanks your comtribute.

siriak (approved):
@imp2002 why not include it? This version is faster and still pretty easy to understand

imp2002:
Yes, this version is really faster. But i dont' think it's more easy to understood. It's a rule that all prime numbers are adjacent to a multiple of `6`, not all people know this. I think the most basic popular approach should be remained and this version add as a optimize function, though the method which is in the repo now is not basic.

siriak:
@fffzlfk could you make it a separate method instead of modifying the existing one?

PR #278, opened 2022-02-22
Change DSU to use both path compression and union by size

er888kh:
This should help with time complexity in big graphs. Specifically, this
can reduce the amortized query time complexity to O(alpha(n)), alpha is
Ackermann function, which grows very slowly compared to log(n) and
doesn't exceed 4 for any n < 10 ^ 600


siriak (approved):
LGTM, thanks!

PR #281, opened 2022-02-28
Add square root by Newtons method

fffzlfk:


er888kh:
I'm not sure whether this is important or not, but big numbers might result in infinite loop (the error would never reach 1e-10 level in f64). It might be useful (or maybe not) to use a relative error threshold or a fixed number of iterations.

ozkanpakdil:
out of curiosity, is this O(logN) ?

er888kh:
> out of curiosity, is this O(logN) ?

Newton's method converges quadratically: when the approximation comes sufficiently close to the actual root, each time its difference with root is squared (assuming the method doesn't break or hit edge cases, which doesn't happen for finding square root of numbers but can happen in more complicated functions).

siriak (approved):
LGTM, thanks!

PR #285, opened 2022-03-04
Generate Nth Prime Number

Hawk453:
An algorithm to generate Nth Prime Number based on Sieve of Eratosthenes.


siriak (approved):

siriak (approved):
LGTM, could you fix the failing check, probably something's up because of merging in master

src/math/nthprime.rs
```diff
@@ -0,0 +1,58 @@
...
+        total_prime = primes[2..].iter().sum();
+        size_factor += 1;
+        s = n * size_factor;
+    }
+
+    count_prime(primes, n).unwrap()
+}
+
+fn get_primes(s: u64) -> Vec<u64> {
+    let mut v: Vec<u64> = vec![1; s as usize];
```
er888kh, line 24: Can you change this to `Vec<bool>`?
It will help with memory usage (and maybe vectorization).

Hawk453: Certainly, will do this.

src/math/nthprime.rs
```diff
@@ -0,0 +1,58 @@
...
+            }
+        }
+    }
+    v
+}
+
+fn count_prime(primes: Vec<u64>, n: u64) -> Option<u64> {
+    let mut counter: u64 = 0;
+    for i in 2..primes.len() {
+        counter += primes.get(i).unwrap();
```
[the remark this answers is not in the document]
Hawk453: clippy recommended get(i) but lets see, I'll do this too.

src/math/nthprime.rs
Hawk453: Utilizing the vec in the function get_nth_prime. The while loop has a use case for the sum of the count.

src/math/nthprime.rs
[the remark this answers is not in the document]
Hawk453: ```the loop variable `i` is used to index `primes` ``` Clippy warning when using primes[i]

Hawk453:
@er888kh Looks like its fine as it is for both suggestions are redundant.

siriak:
The check build / cargo fmt (pull_request)  is failing and the branch is out of date a bit

er888kh:
> @er888kh Looks like its fine as it is for both suggestions are redundant.

Oh I see. So then just do what siriak suggested and call it a day :)

siriak (approved):
Awesome, thanks!

PR #286, opened 2022-03-07
Added new algorithms and tests

DONSIMON92:
* Added Pancake sorting with tests
* Added tests to Quick sort
* Updated two `README.md` (main file with list and with description)


DONSIMON92:
I just repeated changes for @er888kh  to correct clippy warnings in the following files: `src/ciphers/morse_code.rs`, `src/ciphers/polybius.rs`, `src/string/manacher.rs`

PR #290, opened 2022-03-11
Add new algorithms with description

DONSIMON92:
* added pancake sort
* added jump search
* added fibonacci search
* added exponential search
* added description of algorithms in `README.md`
* added algorithms to main list in `README.md`
* fixed the cargo command to the one that starts in CI, so that it would be possible to check for warnings locally


siriak (approved):
Awesome, thanks!

PR #291, opened 2022-03-17
Add Pollard's Rho algorithm

er888kh:
I tried my best to provide a clean and performant implementation, but there may be easy and good optimizations I missed. At first, I was going to implement this using montgomery multiplication to factor up-to 128 bit integers, but then I realized this algorithm sometimes (for example in RSA-like integers) takes very long / doesn't work at all for large integers.

With up-to-date rust-nightly, it takes ~5ms per number on my (ancient!) machine (in unoptimized mode. in optimized mode, each number is factorized in < 0.8ms) for the bigger test case. I wasn't sure if it should be `#[ignore]`'d or not (it takes < 100ms for the whole test case).


er888kh:
Oh, one thing I forgot: I had to implement some RNG for this. I decided Linear Congruence Generator was good enough for this purpose. Should I add LCG as another algorithm to this repository? If yes, should I change something (in this current implementation) before that? And how should the tests be implemented?

PR #293, opened 2022-03-18
Add Fast Fourier Transform

er888kh:
I tried my best to provide a readable and performant implementation, but I couldn't get the bigger test case to run under 1000ms in unoptimized mode (but it runs in ~70ms in release mode). Because of this, I `#[ignore]`'d it. I would be awesome if we can figure out how to make this code run faster in unoptimized mode.


siriak (approved):
LGTM, thanks! Feel free to open another PR if you find some ways to optimize it!

PR #294, opened 2022-03-20
Lowest common ancestor

er888kh:
[New year](https://en.wikipedia.org/wiki/Nowruz), new algorithm!

I implemented LCA algorithm in two ways:
1. Online algorithm using sparse table (also called binary lifting method)
2. Offline algorithm first discovered by Tarjan. The advantage is lower memory consumption and better time complexity / performance. 

Because Tarjan's algorithm required Disjoint Set Union algorithm, I decided to implement that in a separate file. After that, I changed one of the Minimum Spanning Tree algorithms which used DSU to use the new implementation in order to remove duplication.


er888kh:
Hi @siriak, can you not merge this until I add the graph enumeration code to it? The commit will most probably be done by a few hours before this time tommorow.

er888kh:
This should be good now

siriak (approved):
LGTM, thanks!

PR #295, opened 2022-03-20
Create a Pull-Request template

er888kh:
It seems like a good percentage of first-time contributors miss some of the checks that should be done before submitting their PR. Placing a reminder for people to check and double check their code can help reduce PR-to-merge time.

In case I missed something important or some part of this can be written in a better way, please, let me know.


siriak (approved):
Awesome, thanks!

PR #296, opened 2022-03-21
Add hamming distance

saybarsnazlica:
I added an algorithm to find the hamming distance between two strings.

### DIFF
--- a/DIRECTORY.md
+++ b/DIRECTORY.md
@@ -92,3 +92,4 @@
     * [Rabin Karp](https://github.com/TheAlgorithms/Rust/blob/master/src/string/rabin_karp.rs)
     * [Reverse](https://github.com/TheAlgorithms/Rust/blob/master/src/string/reverse.rs)
     * [Z Algorithm](https://github.com/TheAlgorithms/Rust/blob/master/src/string/z_algorithm.rs)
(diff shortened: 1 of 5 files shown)


er888kh:
Thanks for your contribution!
One minor tip to make your code cleaner: In `hamming_distance.rs`, line 10, you could add `unreachable!()` to make it clear that all the possible options were exhausted.

Other than that, I think hou should add `self::hamming_distance::hamming_distance` to the end of `mod.rs` too (by convention.)

saybarsnazlica:
Thanks for the comment. I couldn't get your point about `unreachable!()`. In line 10 the loop reaches the end of the strings so it needs to break.

er888kh:
> Thanks for the comment. I couldn't get your point about `unreachable!()`. In line 10 the loop reaches the end of the strings so it needs to break.

Oh, sorry. I meant line 11 `_ => continue,`
Here is the official [documentation](https://doc.rust-lang.org/std/macro.unreachable.html).

saybarsnazlica:
Oh, I see. I modified my code accordingly.

siriak (dismissed):
LGTM, please fix the failing `cargo fmt` check so that I could merge your changes.

siriak (approved):

PR #297, opened 2022-03-22
Ternary Search Algorithm

mime8:
Added ternary search algorithm implementation.


er888kh:
Can you add one function to each file, such that it takes a closure instead of a list? This way we can do so much more, e.g. find vertex of a parabola or calculate pi using trigonometric functions.

mime8:
> Can you add one function to each file, such that it takes a closure instead of a list? This way we can do so much more, e.g. find vertex of a parabola or calculate pi using trigonometric functions.

Hi, @er888kh . Just to clarify, besides naive approach you also want a version of ternary algorithm for working with unimodal functions, correct?

er888kh:
> Hi, @er888kh . Just to clarify, besides naive approach you also want a version of ternary algorithm for working with unimodal functions, correct?

Yes, that would be great (of course a few more parameters should be added by the user, like two points and some relative minimum acceptable error)

mime8:
> > Hi, @er888kh . Just to clarify, besides naive approach you also want a version of ternary algorithm for working with unimodal functions, correct?
> 
> Yes, that would be great (of course a few more parameters should be added by the user, like two points and some relative minimum acceptable error)

There would be 2 functions actually. For minimum and maximum. Same goes for recursive option.
Any preference for types for start/end bounds? I was thinking about just using float.

er888kh:
It would make sense to use f64

mime8:
> It would make sense to use f64

Didn't see your reply and pushed my changes. Please, tell me if the new changes are what you wanted. I can switch types later.

er888kh:
Don't worry, it's fine. Let's just wait and see if maintainers approve it.

siriak (approved):
Looks good, thanks!

PR #299, opened 2022-03-23
Matrix Operations

b42thomas:
pushing up work for matrix transpose

### DIFF
--- a/src/math/matrix_operations.rs
+++ b/src/math/matrix_operations.rs
@@ -1,0 +1,50 @@
+/*
+  Matrix Operations
+  transpose, sum & difference, scalar multiplication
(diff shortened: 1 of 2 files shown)


er888kh:
Thank you for your contribution. Can you make this implementation generic? Because these functions will be used in many places and in many more algorithms, it would make sense to have a generic implementation.

Another thing to mention: `cargo clippy` will create a warning for using `&Vec<Vec<i32>>` as function input without using its special capabilities in the implementation (at least in `nightly 1.61`.) Please consider changing it to `&[Vec<i32>]` (it doesn't require changing anything else.)

b42thomas:
@er888kh thanks for the feedback, I will work on a generic implementation this week

PR #303, opened 2022-03-29
Add Cipolla algorithm for Quadratic Residue problem

itewqq:
Cipolla algorithm for Quadratic Residue problem

Solving quadratic residue problem:
    x^2 = a (mod p) , p is an odd prime
with O(M*log(n)) time complexity, M depends on the complexity of complex numbers multiplication.

Wikipedia reference: https://en.wikipedia.org/wiki/Cipolla%27s_algorithm
If a is Quadratic Nonresidues, return None; Otherwise return two solutions (x1, x2)


siriak (commented):
Looks good, but could you reuse LCG instead of copying it?

itewqq:
> Looks good, but could you reuse LCG instead of copying it?

No problem.
At that time I was thinking there will be a standalone random generator of this repo in the future, so I counld not determine whether to reuse the one in pollard_rho.rs 😆

er888kh:
> > Looks good, but could you reuse LCG instead of copying it?
> 
> No problem. At that time I was thinking there will be a standalone random generator of this repo in the future, so I counld not determine whether to reuse the one in pollard_rho.rs 😆

The RNG is finally there (#311). While you can use that (when merged), LCG is fine too.

itewqq:
> Merged #311, please either use an rng from #311, or add your LCG in the `random` package and use it, the point is that rng would be separate from the Cipolla algorithm

Thanks for the direction, I modified the code to reuse PCG32 from [#311 ](https://github.com/TheAlgorithms/Rust/pull/311).

er888kh:
> > Merged #311, please either use an rng from #311, or add your LCG in the `random` package and use it, the point is that rng would be separate from the Cipolla algorithm
> 
> Thanks for the direction, I modified the code to reuse PCG32 from [#311 ](https://github.com/TheAlgorithms/Rust/pull/311).

Just one more thing.
It seems like you are using time as seed for the RNG. Can you add a mechanism for the user to seed the RNG?
This has to do with reproducibility for tests.

README.md
```diff
@@ -54,7 +54,11 @@ These are for demonstration purposes only.
 - [X] [Prime number](./src/math/prime_numbers.rs)
 - [x] [Linear Sieve](./src/math/linear_sieve.rs)
 - [x] [Pollard's Rho algorithm](./src/math/pollard_rho.rs)
+<<<<<<< HEAD
+- [x] [Quadratic Residue](./src/math/quadratic_residue.rs)
+=======
 - [x] [Simpson's Rule for Integration](./src/math/simpson_integration.rs)
+>>>>>>> 98ad004c2a223bd1962c35cb60cb8853aa76c509
```
siriak, line 61: Merge went wrong here

itewqq:
> Just one more thing. It seems like you are using time as seed for the RNG. Can you add a mechanism for the user to seed the RNG? This has to do with reproducibility for tests.

Sure, just added the seed paramters for users to use.

siriak (approved):
LGTM, thanks!

PR #305, opened 2022-04-01
Armstrong number 

saybarsnazlica:
An Armstrong (narcissistic) number is a number that is the sum of its digits each raised to the power of the number of digits. This algorithm checks whether a number is an Armstrong number.


siriak (approved):
Looks good, thank you!

PR #306, opened 2022-04-04
Fix `RBTree::find` ref usage

jtescher:
It is unsound to return mutable references from find as you can then trivially produce multiple to the same value.


siriak (approved):
Thanks!

PR #314, opened 2022-04-15
Add Centroid Decomposition

er888kh:
My last test case took 35 secs on my own machine in unoptimized mode and 3.42 secs in optimized mode, so it must definitely be `[#ignore]`'ed. It is also pretty tight, to my own surprise; the theoretical `max_height` isn't too different from the actual result.

Side note: It is so satisfying to see a lot of the "infrastructure" we have added being used at the same time: `PCG32`, `prufer_code`, and `enumerate_graph`.

Edit: After some investigation, it seems like the code spends ~70% of its time in the test case to generate the tree, which makes sense as the Prufer code implementation cannot directly output the graph representation used here.


siriak (approved):
LGTM, thanks!

Issue #319, opened 2022-04-22
Bug in `string/aho_corasick` algorithm


guopenghui:
https://github.com/TheAlgorithms/Rust/blob/a5cf37c397d097305f52736d6305d22f22b7b82a/src/string/aho_corasick.rs#L80

It will cause computing overflow when pattern appear at head of string.

If `i - len + 1 == 0`,  `i - len` will be `-1` first,  then panic.

Change to `i + 1 - len` will solve this.

siriak:
Thanks for reporting, could you make a PR with the proposed changes?

PR #322, opened 2022-05-04
Prevent potential computing overflow in aho_corasick.rs.

merelymyself:
As per issue #319, there was a chance of computing overflow due to the order of operations. This minor commit should fix that.


siriak (approved):
LGTM, thanks!

PR #330, opened 2022-05-18
add assymentric test to string reverse

materoy:


src/string/reverse.rs
```diff
@@ -11,6 +11,11 @@ mod tests {
         assert_eq!(reverse("racecar"), "racecar");
     }
 
+    #[test]
+    fn test_assymentric() {
```
siriak, line 15:
```suggestion
    fn test_assymetric() {
```

siriak (approved):
Thanks!

Issue #333, opened 2022-05-23
Miller-Rabin and 0


casperes1996:
In the algorithm for the Miller-Rabin primality test, 0 is reported as a prime. 0 should be special-cased by an initial "if 0 return false;" check. One can argue 0 is just invalid input for the algorithm, but I would then argue at least throwing an assert error on 0 input is preferable so it becomes clear to anyone reading or potentially copying the algorithm for primality testing that 0 is not a prime but not accounted for in the current algorithm logic.

siriak:
Done

PR #334, opened 2022-05-23
Add 0 check for primality test as per issue post #333

casperes1996:
As argued in Issue #333 , this should help make it clear 0 is not a prime. Other options can be discussed like using optionals, Result, etc. Or even just a comment saying 0 is invalid input and not handling it outside of that


siriak (approved):
Agree, thanks for contributing!

PR #339, opened 2022-06-07
Fixed binary search for descending order elements

RajMazumder18110:
Screenshot 2022-06-07 at 11 44 35 AM


siriak (approved):
Looks good, thanks! The implementation trick for determining sorting direction based on the first and the last elements is cool :)

PR #340, opened 2022-06-08
added interpolation_search

Yazeed1s:
## Description
Hi there, 

I implemented interpolation search algorithm and followed all `COUNTRIBUTING.md`'s guidelines. The algorithm itself is simple, it is an improvement over `binary search`. 

Unlike binary search, interpolation search does not go to the middle element to check, instead, it jumps to the side where the target being searched is located in. 

My implementation does not take generic data types, it only takes `i32`. However, I tried using `T` instead, but the borrow-checker did not allow me at all.
## Type of change

## Checklist:


siriak (approved):
Looks good, thanks!

Yazeed1s:
My pleasure

PR #343, opened 2022-06-28
Fixed typo in comment of binary_search_tree.rs

Soakeson:
I found a typo in a binary_search_tree.rs comment and thought I'd try my hand at fixing it. This is the first time I've ever tried contributing to something, any feedback on what to do better in the future would be greatly appreciated.

### DIFF
--- a/src/data_structures/binary_search_tree.rs
+++ b/src/data_structures/binary_search_tree.rs
@@ -34,7 +34,7 @@ where
         }
     }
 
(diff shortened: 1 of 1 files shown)


alexanderwid:
iff means if and only if so it's not necessarily a typo.

siriak:
Agree with @alexanderwid, this is not a typo

Soakeson:
That's super interesting! Thanks for taking the time to comment!

PR #345, opened 2022-07-03
Add 2-SAT solver

er888kh:
I'm not sure if there is a way to decrease stack usage in `strongly_connected_components` dramatically or if there is a better solution for running `big_test` without running it in another thread. I tried increasing maximum stack usage limit in my environment, but it seems like something from within `cargo test` itself is limiting the maximum stack usage for test cases.


imp2002 (approved):
LGTM! Thanks.

PR #351, opened 2022-07-31
Add a generic HMAC struct and a Hashing trait

er888kh:
Hi,
This is a draft of what I think our cryptography infrastructure should look like. We should at least try to have a trait for various cryptographic hashing function that we implement in the future. Also I tried to implement HMAC based on [RFC 2104](https://www.rfc-editor.org/rfc/rfc2104) using the proposed trait. I am looking for feedback on the Hashing trait and whether the most important functionality is there.

I really wanted to have a demo for HMAC using the already implemented `sha256` function, but after having a closer look, I realized this function has some implementation flaws ([here](https://github.com/TheAlgorithms/Rust/blob/83ad008c24d9d264039dabda2375800da0f38d5b/src/ciphers/sha256.rs#L122) for example) that slow it down considerably. As such I need more time and the feedback from the Hashing trait before I can test this HMAC implementation.

### DIFF
--- a/src/ciphers/hashing_traits.rs
+++ b/src/ciphers/hashing_traits.rs
@@ -1,68 +1,72 @@
 pub trait Hasher<const DIGEST_BYTES: usize> {
-	/// return a new instance with default parameters
-	fn new_default() -> Self;
+    /// return a new instance with default parameters
+    fn new_default() -> Self;
 
-	/// Add new data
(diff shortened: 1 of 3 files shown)


siriak:
I like your approach, @imp2002 what do you think?

siriak (approved):
Looks good, thanks for contributing!

PR #353, opened 2022-08-19
Optimize graph breadth_first_search

aggresss:
On `insert` operation, If the set did have the value present, `false` is returned. So there is no need to use `contains` operation.


siriak (commented):
Thanks for suggesting, but with the proposed changes the algorithm loses expressiveness because it relies on a side effect of operation and not its primary purpose. For educational purposes, it's better to have these 2 operations clearly separated.

PR #354, opened 2022-08-22
fix: Make xor output Vec<u8>

Ongy:
String is documented as utf8 encoded text.
This can contain multibytes. The cast of char to u8 is lossy.

test_multi_byte fails before the API change.
The other new tests also test common encoding issues.
The don't fail on the old code, but the intermediaries are invalid.

This could also be done with e.g. `Box<[u8]>`, but String is growable, so `Vec<u8>` should be fine.

This changes the API, but since the old one necessarily violates documented type properties, IMO it needs to be done.


siriak (commented):
The changes look good, but a PR check fails for unrelated reasons. Could you please fix it? Shouldn't take long

siriak (approved):
Thanks a lot!

PR #355, opened 2022-08-22
Added faster_perfect_numbers and mersenne_primes

Utshaan:
A function that evaluates perfect numbers till a certain limit using the Lucas Lehmer test algorithm making it substantially faster.

As this algorithm requires the use of mersenne primes, a module for the same was added with.


siriak (approved):
Looks good, thanks!

PR #356, opened 2022-08-22
Remove redundant `sqrt()` calls in closest points

kinkard:


src/geometry/closest_points.rs
```diff
@@ -42,12 +42,12 @@ fn closest_points_aux(
 
     if n <= 3 {
         // bruteforce
-        let mut min = dist(&points[0], &points[1]);
+        let mut min = sqr_dist(&points[0], &points[1]);
         let mut pair = (points[0], points[1]);
 
-        for i in 0..n {
+        for i in 1..n {
```
siriak, line 48: Why change indexing?

kinkard: Because we already computed values for `0` + `1` indexes in a few lines above
```rust
        let mut min = sqr_dist(&points[0], &points[1]);
        let mut pair = (points[0], points[1]);
```

What is the point to recalculate it again for this pair of indexes?

src/geometry/closest_points.rs
```diff
@@ -42,12 +42,12 @@ fn closest_points_aux(
 
     if n <= 3 {
```
siriak, line 43: Shouldn't this be `n >= 3`?

siriak (commented):
Your changes look good, could you add a test or two for cases with more than 3 points?

kinkard:
> Your changes look good, could you add a test or two for cases with more than 3 points?

This algorithm already has exhaustive tests, including testing cases with more than 3 points:
https://github.com/TheAlgorithms/Rust/blob/66e19f0c77e591df3f8c7984de08767fc28d5afb/src/geometry/closest_points.rs#L155-L220

src/geometry/closest_points.rs
kinkard: Nope, if `n >= 3` the simple bruteforce is too expensive.
It's a recursive algorithm where all array of points is recursive checked and the tail of the recursion is the case when points less or equal than 3.

siriak (approved):
Looks good, thanks!

PR #363, opened 2022-08-24
Implement salsa streaming cipher

er888kh:
A basic implementation of salsa streaming cipher. Also fixed a new warning from clippy in `src/graph/depth_first_search_tic_tac_toe.rs`


siriak (approved):
Looks good, thanks!

PR #364, opened 2022-08-26
linked_list: insert_at_ith updates all pointers

NathanYee:
Inserting at index `i` does not update the `prev` pointer of the node that used to be at index `i`. Currently, when inserting many nodes at index `1`, all the node's `prev` pointer points to the `head` of the list. Deleting any of the nodes in the middle will delete the `head` of the list. A test as been added for this case.

PS
Love the repository. It is super useful!


siriak (changes requested):
Please fix the build check `cargo fmt`, everything else looks good

NathanYee:
Good catch. Should be good to go now.

siriak (approved):
Looks good, thanks!

PR #365, opened 2022-08-27
Improve Two Sum example

Overflow0xFFFF:
Changes:
* Add rustdoc
* Add comments describing the solution
* Use a more correct return type
* Remove the use of unwrap
* Add a test case for when a sum isn't found in the list

These changes build on the solution already submitted. The addition of rustdoc and comments are meant to help learners who are approaching the language for the first time.

The Two Sum problem itself asks that we only return two indices from within the list of numbers, so it's more natural to use a Tuple of `usize`s here than a `Vec<i32>`! This means we can eliminate the use of `unwrap` altogether and have some slightly more idiomatic Rust code.

It's worth noting that I couldn't find a way to improve the original solution, so all credit goes to the original author for a job well done! :smile:


src/general/two_sum.rs
```diff
@@ -1,24 +1,51 @@
...
-pub fn two_sum(nums: Vec<i32>, target: i32) -> Vec<i32> {
-    let mut hash_map: HashMap<i32, i32> = HashMap::new();
-
-    for (i, item) in nums.iter().enumerate() {
-        match hash_map.get(&(target - item)) {
-            Some(value) => {
-                return vec![i.try_into().unwrap(), *value];
+    for (i, current_value) in nums.iter().enumerate() {
+        match distance_table.get(&(target - current_value)) {
+            Some(j) => {
```
dxbednarczyk, line 38: You can avoid the use of brackets here:

```rust
// snip
Some(j) => return Some((i, *j)),
_ => distance_table.insert(*current_value, i),
```
although it is up to personal preference at that point.

Overflow0xFFFF: I think it's worth changing. :slightly_smiling_face:

dxbednarczyk (approved):

src/general/two_sum.rs
dxbednarczyk: It won't let me resolve the conversation after your force push although everything seems fine

siriak (approved):
Looks good, thanks!

PR #369, opened 2022-09-04
Add run length encoding

kou-sia:
I implement&test run-length-encoding.


kou-sia:
@siriak @imp2002 
I would like to add an algorithm that does not yet exist in this repository.

siriak (commented):
The code looks good, but could you also add a decoder? It seems that having only an encoder has no point because it can't be meaningfully used

kou-sia:
@siriak 
Thanks for the review.
You are absolutely right to make that point.
I will also add decoder.

kou-sia:
I implemented the decode function.
And I implemented this function as "pub fn run_length_decoding".
In this case, do I need to do the following?
1. add explanation to "src/string/README.md".
3. add a decode function in "src/string/mod.rs".

kou-sia:
Checking a similar case, 'burrows_wheeler_transform', the respective function was added to 'src/script/mod.rs', so it was added in the same way.

siriak (approved):
Looks good, thanks!

kou-sia:
@siriak 
So Sorry.
I forgot do "cargo fmt & cargo clippy".
I did it. https://github.com/TheAlgorithms/Rust/pull/369/commits/07cd25a278936618db48245ffeff43a8a051bff7

siriak (approved):

PR #371, opened 2022-09-05
Add Poly1305

er888kh:
Implementation for poly1305 Message Authentication Code.

This is the initial effort to add big-integer arithmetic as an optional (but enabled by default) feature. Here I used `num-bigint` because it seemed to be well-maintained and well-written. The other candidates were based on `gmp` C library and were bindings on top of that. While using `gmp` had its own advantages (mainly performance and decades of testing), it wasn't selected due to it being hard to compile on Windows systems and we want the code in this repository to be as portable as possible.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,13 @@
 [package]
 name = "the_algorithms_rust"
+edition = "2021"
 version = "0.1.0"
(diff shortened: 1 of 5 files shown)


siriak (approved):
Looks good, thanks!

PR #373, opened 2022-09-06
Quick Sort Tests

ahmed-deftoner:
The quick sort code was failing when the slice was empty. There weren't any tests in the file as well. I tweaked the code to solve the empty array issue and added some tests along with comments :)


siriak (dismissed):
Great, thanks! Could you run `cargo fmt` to fix a PR check?

siriak (approved):
Looks good, thanks!

PR #378, opened 2022-09-20
Chinese remainder theorem

GentBinaku:


siriak (approved):
Looks good, thanks!

er888kh:
Seems like I'm a little too late.

Is there any chance @GentBinaku you can change this to work with an arbitrary number of u64 numbers?

Also, can you make another version with big number support?

In an ideal world the two versions can be made using a simple generic implementation. But that's only if we won't suffer from weird syntax or reduced performance.

It's okay if you can't do both though. I will have a good amount of time to work on this in the first week of October.

GentBinaku:
@er888kh 
Yes I can

GentBinaku:
Hi @er888kh and @siriak, I have an interview coming up and I am going to study for it, so after that, I will pick this up.

er888kh:
Good Luck @GentBinaku!
I hope soon enough I can get back too. Right now the situation doesn't look too good here in Iran. To use internet, we have to stay awake all night and jump through all sorts of hoops to get maybe 50kb/s bandwidth with 50% packet loss rate. Unless the police starts killing people in thousands, like in 2019, I'm not sure whether things will start to look better anytime soon.

siriak:
@er888kh good luck and take care of yourself! I'm from Ukraine and things are not looking up here too. It has been a difficult year for everybody and I don't think struggles end any time soon, but I'm sure next year will be much better for all of us!

PR #379, opened 2022-09-21
add utf8 string support for aho corasick

ptrsxu:
The former Aho Corasick code fails on non-ascii utf8 strings. This pr fix the problem.


siriak (changes requested):
The code looks good, but please fix the build check

ptrsxu:
> The code looks good, but please fix the build check

ok, updated.

siriak (approved):
Looks good, thanks!

PR #380, opened 2022-09-21
ADD TEST : burrows_wheeler_transform

kou-sia:
add TEST : burrows_wheeler_transform's test for ensure function stand-alone legitimacy
&
fix : variable name IN run_length_decoding

### DIFF
--- a/src/string/run_length_encoding.rs
+++ b/src/string/run_length_encoding.rs
@@ -31,7 +31,7 @@ pub fn run_length_decoding(target: String) -> String {
     }
 
     let mut character_count: String = String::new();
-    let mut encoded_target = String::new();
+    let mut decoded_target = String::new();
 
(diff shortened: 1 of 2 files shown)


kou-sia:
@siriak @imp2002 
This is because if the function and the inverse function have errors in the same place, the test of the composite function alone cannot guarantee the validity of the function.

src/string/run_length_encoding.rs
```diff
@@ -31,15 +31,15 @@ pub fn run_length_decoding(target: String) -> String {
     }
 
     let mut character_count: String = String::new();
-    let mut encoded_target = String::new();
+    let mut decoded_target = String::new();
```
kou-sia, line 34: rename

siriak (approved):
Looks good, thanks!

PR #384, opened 2022-09-24
add sleep_sort

kou-sia:
I want to add sleep-sort.

### DIFF
--- a/src/sorting/README.md
+++ b/src/sorting/README.md
@@ -180,16 +180,14 @@ __Properties__
 From [Wikipedia][tim-wiki]:  Timsort is a hybrid stable sorting algorithm, derived from merge sort and insertion sort, designed to perform well on many kinds of real-world data. It was implemented by Tim Peters in 2002 for use in the Python programming language. The algorithm finds subsequences of the data that are already ordered (runs) and uses them to sort the remainder more efficiently. This is done by merging runs until certain criteria are fulfilled. Timsort has been Python's standard sorting algorithm since version 2.3. It is also used to sort arrays of non-primitive type in Java SE 7, on the Android platform, in GNU Octave, on V8, Swift, and Rust.
 
 __Properties__
-* Worst-case performance O(n log n)
-* Best-case performance	O(n)
+* Worst-case performance O(max element size(ms))
+* Best-case performance	O(max element size(ms))
 
 ### [Sleep](./sleep_sort.rs)
 ![alt text][sleep-image]
 
 From [Wikipedia][bucket-sort-wiki]: This is an idea that was originally posted on the message board 4chan, replacing the bucket in bucket sort with time instead of memory space.
 It is actually possible to sort by "maximum of all elements x unit time to sleep". The only case where this would be useful would be in examples.
-
-
 
 [bogo-wiki]: https://en.wikipedia.org/wiki/Bogosort
 [bogo-image]: https://upload.wikimedia.org/wikipedia/commons/7/7b/Bogo_sort_animation.gif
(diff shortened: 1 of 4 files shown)


kou-sia:
sleep sort is a half-joking sort method, but there are opportunities to see it on the Internet in the process of learning programming. And I think there is a certain value in being able to actually see an implementation in the Rust language.

src/sorting/README.md
```diff
@@ -180,8 +180,14 @@ __Properties__
...
 __Properties__
-* Worst-case performance O(n log n)
-* Best-case performance	O(n)
+* Worst-case performance O(max element size(ms))
+* Best-case performance	O(max element size(ms))
+
+### [Sleep](./sleep_sort.rs)
+![alt text][sleep-image]
+
+From [Wikipedia][bucket-sort-wiki]: This is an idea that was originally posted on the message board 4chan, replacing the bucket in bucket sort with time instead of memory space.
```
kou-sia, line 189: Since we could not confirm wikipedia in English, I refer to the section described in Japanese.
(If necessary, I would be willing to translate it into English and create a wikipedia page.)

src/sorting/README.md
```diff
@@ -235,3 +241,6 @@ __Properties__
 
 [comb-sort]: https://upload.wikimedia.org/wikipedia/commons/4/46/Comb_sort_demo.gif
 [comb-sort-wiki]: https://en.wikipedia.org/wiki/Comb_sort
+
+[sleep-sort]: <no image>
+[sleep-sort-wiki]https://ja.m.wikipedia.org/wiki/バケットソート#.E3.82.B9.E3.83.AA.E3.83.BC.E3.83.97.E3.82.BD.E3.83.BC.E3.83.88
```
kou-sia, line 246: Since we could not confirm wikipedia in English, we refer to the section described in Japanese.
(If necessary, I would be willing to translate it into English and create a wikipedia page.)

src/sorting/sleep_sort.rs
```diff
@@ -0,0 +1,88 @@
...
+use std::time::Duration;
+
+pub fn sleep_sort(vec: &[usize]) -> Vec<usize> {
+    let len = vec.len();
+    let (tx, rx) = mpsc::channel();
+
+    for &x in vec.iter() {
+        let tx: mpsc::Sender<usize> = tx.clone();
+        thread::spawn(move || {
+            thread::sleep(Duration::from_millis((10 * x) as u64));
```
kou-sia, line 12: I multiply it by 10 for safety.

archuser555:
can you explain what did this do

kou-sia:
@archuser555 
Hi, OK! The following processes are performed!
1. receive an array
2-A. asynchronous parallel processing of its entire elements ```for &x in vec.iter()```
2-B. processing is to sleep for the size of the elements, and ```tx.send()``` is performed from the element with the smallest size.
3. perform a for-loop for the number of elements, and push them to the array in order by ```rx.recv().unwrap()```.
4. the array is sorted after push

ex. [5, 1, 3]
1. receive [5, 1, 3]
2-A. parallel process "5" & "1" & "3"
2-B. about "5" push it after 50ms 
        about "1" push it after 10ms
        about "3" push it after 30ms
3. So, "1", "3" and "5" are stored  in ```turn.rx.recv()``` in turn.
4. push it to sorted_list

src/sorting/README.md
siriak: Having a page in English would be nice. I can merge it with Japanese reference, and then you translate it and make a PR with updated English reference when you have time and energy to do so.

PR #386, opened 2022-09-26
Adding Bakhshali Sqrt Algorithm

archuser555:
This Is A Technique To Calculate Square Root, I Write It In Rust, And Yeah, It's Indian Old Way But Work's

### DIFF
--- a/src/math/bakhshali_sqrt.rs
+++ b/src/math/bakhshali_sqrt.rs
@@ -1,0 +1,44 @@
+#[allow(path_statements, unused_must_use)]
+//Bakhshali Sqrt Algorithm
+
+//support f32 so please use 10.0 instead of 10
+//this is an ancient indian way to calculate square root, so don't wait very exact result
+pub fn bakhshali_sqrt(x: f32) -> f32 {
+    let mut res: f32;
+    let mut a: f32;
+    let mut b: f32;
+
+    if (x == 0.0) || (x == 1.0) {
+        return x;
+    } else if x < 0.0 {
+        return f32::NAN;
+    }
+
+    res = x * 0.25;
+
+    loop {
+        let pre_res: f32 = res;
+
+        a = (x - (res * res)) / (2.0 * res);
+        b = res + a;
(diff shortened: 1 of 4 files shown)


src/math/bakhshali_sqrt.rs
```diff
@@ -0,0 +1,44 @@
+#[allow(path_statements, unused_must_use)]
```
merelymyself, line 1: are these necessary?

archuser555: Kinda Of, Deleting Them Cause Some Warnings

archuser555: After Check I Found That They Actually Do Nothing :-)

src/math/bakhshali_sqrt.rs
```diff
@@ -0,0 +1,44 @@
...
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    pub fn test_bakhshali_sqrt() {
+        // let's try get the sqrt of 10
+        let a: f32 = bakhshali_sqrt(9.0);
+        assert_eq!(a, 3.0); //Sqrt Of 9 Is 3
+    }
```
siriak, line 43: Please add more tests with 0, negative numbers, 0 to 1 numbers, non-perfect-square integer numbers

siriak (changes requested):

src/math/bakhshali_sqrt.rs
archuser555: Ok, Gonna Add Some!

archuser555: Done :-)

siriak (commented):
The code looks good, but fix the build check please

zim0369:
If OP is busy then I could make a PR in case this one is going to be abandoned. Do I?

archuser555:
idk what's the prob :-)
ok do it! but mention me!
@zim0369

PR #394, opened 2022-10-01
Added fisher yates shuffle

jerr-it:


merelymyself:
As a rule, we try to reduce the number of crates we have as dependencies. Is there an implementation without rand?

jerr-it:
I don't think it's possible to implement the Fisher-Yates-Shuffle without randomness. But using the random generator in math/random.rs should work.

siriak (approved):
Looks good, thanks!

Issue #375, opened 2022-09-13
Linting Errors and Warnings


Riceman2000:
There are a few linting errors and warnings currently, especially in `./src/data_structures` and `./src/graph`

Examples:

<details>
<summary> parameter is only used in recursion --at--> src\data_structures\b_tree.rs:104:39 </summary>

```
error: parameter is only used in recursion
   --> src\data_structures\b_tree.rs:104:39
    |
104 |     fn traverse_node<T: Ord + Debug>(&self, node: &Node<T>, depth: usize) {
    |                                       ^^^^
    |
    = note: `-D clippy::only-used-in-recursion` implied by `-D warnings`
note: parameter used here
   --> src\data_structures\b_tree.rs:110:17
    |
110 |                 self.traverse_node(&node.children[index], _depth);
    |                 ^^^^
...
115 |             self.traverse_node(node.children.last().unwrap(), _depth);
    |             ^^^^
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#only_used_in_recursion
```

</details>

<details>
<summary> parameter is only used in recursion --at--> src\data_structures\linked_list.rs:185:26 </summary>

```
error: parameter is only used in recursion
   --> src\data_structures\linked_list.rs:185:26
    |
185 |     fn get_ith_node(&mut self, node: Option<NonNull<Node<T>>>, index: i32) -> Option<&T> {
    |                          ^^^^
    |
note: parameter used here
   --> src\data_structures\linked_list.rs:190:22
    |
190 |                 _ => self.get_ith_node(unsafe { (*next_ptr.as_ptr()).next }, index - 1),
    |                      ^^^^
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#only_used_in_recursion
```

</details>

<details>
<summary> parameter is only used in recursion --at--> src\general\huffman_encoding.rs:63:17 </summary>

```
error: parameter is only used in recursion
  --> src\general\huffman_encoding.rs:46:10
   |
46 |         &self,
   |          ^^^^
   |
note: parameter used here
  --> src\general\huffman_encoding.rs:63:17
   |
63 |                 self.get_alphabet(height + 1, path, node.left.as_ref().unwrap(), map);
   |                 ^^^^
64 |                 self.get_alphabet(
   |                 ^^^^
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#only_used_in_recursion
```

</details>

<details>
[...]

PR #395, opened 2022-10-02
Fix all clippy warnings

Arjun31415:


Arjun31415:
Fix #375

merelymyself:
Is it possible to have solutions that don’t involve #[allow(..)]?

Arjun31415:
> 

alright ill do that

PR #401, opened 2022-10-05
Add Dutch National Flag Sort

CenTdemeern1:
Woo, happy Hacktoberfest!
I felt like doing my first code contribution in Rust, and since I'm not sure if I am doing this right, I might need some help on this.
TheAlgorithms already has implementations in Python and Java for this algorithm, so I thought I'd do one in Rust!


siriak (approved):
Looks good, thanks!

PR #402, opened 2022-10-06
Added collatz_sequence.rs

Utshaan:
Added a function to calculate the sequence for any number, with respect to the collatz conjecture.


siriak (approved):
Looks good, thanks!

PR #413, opened 2022-10-26
Chinese remainder theorem

GentBinaku:


src/math/chinese_remainder_theorem.rs
```diff
@@ -1,35 +1,97 @@
...
 }
 
 #[cfg(test)]
 mod tests {
     use super::*;
-
     #[test]
     fn basic() {
         assert_eq!(chinese_remainder_theorem(&[3, 5, 7], &[2, 3, 1]), Some(5));
-        assert_eq!(chinese_remainder_theorem(&[1, 4, 6], &[3, 5, 7]), Some(34));
```
siriak, line 31: Why have you removed that?

PR #414, opened 2022-10-26
add Duval algorithm

a1396537376:
add Duval algorithm, Reference impl: [Duval algorithm](https://cp-algorithms.com/string/lyndon_factorization.html#duval-algorithm).
if anything need to change please tell me.


siriak (approved):
Looks good, thanks!

Issue #423, opened 2022-11-15
Bug in math/ceil


CatchemAL:
**Issue**
`ceil` function is only accurate to 1dp. For cases where `n < num < n + 0.05`, it will incorrectly round down.

**Demonstrating the issue**
```
    #[test]
    fn positive_decimal() {
        let num = 3.01;
        assert_eq!(ceil(num), num.ceil());
    }
```

PR #424, opened 2022-11-15
fix math/ceil function for issue #423

CatchemAL:
`ceil` function is only accurate to 1dp. For cases where `n < num < n + 0.05`, it will incorrectly round down.


siriak (changes requested):
Please add new tests instead of mutating existing

CatchemAL:
Done

siriak (approved):
Looks good, thanks!

PR #350, opened 2022-07-29
#349: improve performance of insertion sort

SteveLauC:
Closes #349 .

This PR is for #349 :)

PR #425, opened 2022-11-16
docs: add more docs for insertion sort

SteveLauC:
This PR is only about documentation, no actual code changes are involved.

In #350 I improved the implementation of the insertion sort. Today I reviewed my impl, finding it is poorly documented and can be confusing for those algorithm newbies especially compared to the previous implementation. This PR adds more explanatory comments to make it easier to understand :)


siriak:
Not needed, thanks!

PR #428, opened 2022-11-19
fix: floyd warshall d[i][i] == 0 case

wingkwong:
In the existing logic, if $u == v$, $v$ is considered as unreachable, while the distance (or cost) should be $0$ instead. This PR is to add that logic and corresponding test cases.


siriak:
@imp2002 let me know what you think about this PR

siriak (approved):
Looks good, thanks!

PR #431, opened 2022-11-24
Simplify `sleep_sort.rs`

mcthesw:
use `while let` and `drop(tx)` to simplify the code


siriak:
@imp2002 what do you think about the proposed change?

imp2002:
unnessary

PR #433, opened 2022-12-06
feat: added abs implementation 

SpiderMath:
This PR aims to add the implementation for the [Abs function](https://the-algorithms.com/algorithm/abs).


siriak:
@imp2002 what do you think about adding this method? I'd say that it's too simple to be mentioned here

PR #435, opened 2022-12-11
Improve `Insertion sort` readability while maintaining performance

### LINKED ISSUE #434: Improve `Insertion sort` readability while maintaining performance
The current [insertion sort](https://github.com/TheAlgorithms/Rust/blob/master/src/sorting/insertion_sort.rs) algorithm differs greatly from the commonly implemented one, and could be improved in readability while maintaining performance

By using the same logic presented in #349 ("move items back instead of swapping them"), we can implement a more compact and simple algorithm, as shown below:

```rust
pub fn insertion_sort<T: Ord + Copy>(arr: &mut [T]) {
    for i in 1..arr.len() {
        let mut j = i;
        let cur = arr[i];

        while j > 0 && cur < arr[j - 1] {
            arr[j] = arr[j - 1]; // move items back instead of swapping them
            j -= 1;
        }

        arr[j] = cur;
    }
}
```

For performance analysis, let's run the algorithm above and the currently implemented one for an input of size 100_000:

```rust
use std::cmp;
use std::time::Instant;
use rand::random;

// The current algorithm
fn insertion_sort_current<T>(arr: &mut [T])
where
    T: cmp::PartialOrd + Copy,
{
    // ...
}

// The proposed algorithm 
fn insertion_sort_improved<T: Ord + Copy>(arr: &mut [T]) {
    // ...
}

fn run_current(arr: &mut [u64]) {
    let start = Instant::now();
    insertion_sort_current(arr);
    let duration = start.elapsed();

    println!("Time elapsed in current insertion sort is: {:?}", duration);
}

fn run_improved(arr: &mut [u64]) {
    let start = Instant::now();
    insertion_sort_improved(arr);
    let duration = start.elapsed();

    println!("Time elapsed in improved insertion sort is: {:?}", duration);
}

fn main() {
    const ARRAY_SIZE: usize = 100_000;

    let mut arr_1: [u64; ARRAY_SIZE] = [0; ARRAY_SIZE]; 

    for i in 0..ARRAY_SIZE {
        arr_1[i] = random();
    }

    let mut arr_2 = arr_1.clone();

    run_current(&mut arr_1);
    run_improved(&mut arr_2);
}
```

The output:
```
$ cargo run
Time elapsed in current insertion sort is: 16.033703849s
Time elapsed in improved insertion sort is: 15.836348153s
```

As you can see, the performance between the two algorithms is practically the same, but the readability improves considerably

willpinha:
Closes #434

### DIFF
--- a/src/sorting/insertion_sort.rs
+++ b/src/sorting/insertion_sort.rs
@@ -1,32 +1,18 @@
-use std::cmp;
-
 /// Sorts a mutable slice using in-place insertion sort algorithm.
 ///
 /// Time complexity is `O(n^2)`, where `n` is the number of elements.
 /// Space complexity is `O(1)` as it sorts elements in-place.
-pub fn insertion_sort<T>(arr: &mut [T])
-where
-    T: cmp::PartialOrd + Copy,
-{
+pub fn insertion_sort<T: Ord + Copy>(arr: &mut [T]) {
     for i in 1..arr.len() {
+        let mut j = i;
         let cur = arr[i];
-        let mut j = i - 1;
 
-        while arr[j] > cur {
-            arr[j + 1] = arr[j];
-            if j == 0 {
-                break;
-            }
+        while j > 0 && cur < arr[j - 1] {
+            arr[j] = arr[j - 1];
             j -= 1;
         }
 
-        // we exit the loop from that break statement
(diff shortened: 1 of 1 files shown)


siriak (approved):
Looks good, thanks!

PR #440, opened 2023-01-17
Fix Ciphers README

LingkKang:
Just find that _Transposition_ in `ciphers` section was implemented but not linked in README.


siriak (approved):
Looks good, thanks!

PR #444, opened 2023-01-29
resolved clippy::uninlined-format-args warnings for math/interest.rs

Rust-Trends:
resolved clippy::uninlined-format-args warnings for math/interest.rs


siriak (approved):
Looks good, thanks!

PR #450, opened 2023-02-07
feat: Add Euler problem 6 resolution

domix80:


domix80:
Suggestion: Create a subfolder in math, Project_Euler where we can add the Euler problem resolutions:
https://projecteuler.net/archives

PR #451, opened 2023-02-09
Add Fast Inverse Square Root

CenTdemeern1:
Adds Quake 3 Arena's Fast Inverse Square Root algorithm.
Tries to stay true to Quake 3 Arena's implementation, which disregards rustfmt. Is this allowed? (As a possible exception)

### DIFF
--- a/src/math/fast_inverse_square_root.rs
+++ b/src/math/fast_inverse_square_root.rs
@@ -1,0 +1,72 @@
+/*
+A direct Rust translation of the implementation of the Fast Inverse Square Root from Quake 3 Arena found on Wikipedia.
+Includes original code comments from Quake 3 Arena by id Software.
+Doesn't care about the sign of the input as the Quake 3 Arena implementation doesn't either.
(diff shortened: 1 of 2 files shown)


siriak (changes requested):
I see the rationale behind including original formatting and comments, but the purpose of this repository is a collection of algorithms for educational purposes. For both learning algorithms and learning coding best practices. What this repository is not for is preservation of historical algorithm implementations.
Having said that, I don't think including nonstandard formatting and profanity are among industry standards. Please remove historical trivia and rewrite the algorithm with best practices in mind, thanks in advance!
P.S. Having a reference to a Wikipedia page which would also provide some historical context is acceptable and even desirable in my opinion.

CenTdemeern1:
The reason I kept the formatting and stuff the same is so that someone who is learning Rust can compare the code to its original C counterpart. If this is a problem I'll rewrite it.

siriak:
Yes, please rewrite it

PR #460, opened 2023-02-18
Feat: Add Bitonic Sort

domix80:
# Pull Request Template

## Description

## Type of change

## Checklist:


siriak (approved):
Looks good, thanks!

PR #463, opened 2023-02-27
fix: clippy::unit_cmp warning

BaiXinKe:
## Description

fix: clippy::unit_cmp warning: Since the sort() method of 'vec' does not have a return value, calling it directly from an assert statement will always evaluate to true, triggering a "unit_cmp" error reported by Clippy.

## Type of change

- [√] Bug fix (non-breaking change which fixes an issue)

## Checklist:

- [√] I ran bellow commands using the latest version of **rust nightly**.
- [√] I ran `cargo clippy --all -- -D warning` just before my last commit and fixed any issue that was found.
- [√] I ran `cargo fmt` just before my last commit.
- [√] I ran `cargo test` just before my last commit and all tests passed.
- [√] I checked `COUNTRIBUTING.md` and my code follows its guidelines.

### DIFF
--- a/src/string/autocomplete_using_trie.rs
+++ b/src/string/autocomplete_using_trie.rs
@@ -111,14 +111,19 @@ mod tests {
         let prefix = "app".to_owned();
         let mut auto_completed_words = auto_complete.find_words(prefix);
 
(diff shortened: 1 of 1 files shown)


siriak (approved):
Looks good, thanks!

PR #464, opened 2023-02-28
feat: adding navigation module

lucioroadtoglory:
## Description

Algorithms about navigation:
- [x] haversine: distance between two points.
- [x] bearing: angle between two points. (Useful to show the direction of a marker, like A to B (A -> B)).

## Type of change

## Checklist:


siriak (approved):
Looks good, thanks!

PR #468, opened 2023-03-10
Clippy

qsantos:
## Description

I'd like to contribute the A* algorithm to the project, but I noticed many clippy warnings. This PR addresses most of them, and ignores 2. You can look commit by commit for each type of lint rule. I've seen that there is [a workflow](https://github.com/TheAlgorithms/Rust/actions/runs/4362844991/jobs/7628222154) running `cargo clippy`, but maybe it's using an older version?

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)

## Checklist:

- [x] I ran bellow commands using the latest version of **rust nightly**.
- [x] I ran `cargo clippy --all -- -D warning` just before my last commit and fixed any issue that was found. (**NOTE:** it's `warnings` with an s (as in the CONTRIBUTING.md), `--all` looks obsolete)
- [x] I ran `cargo fmt` just before my last commit.
- [x] I ran `cargo test` just before my last commit and all tests passed.
- [x] I checked `COUNTRIBUTING.md` and my code follows its guidelines. (**NOTE:** it's `CONTRIBUTING.md`)

PR #469, opened 2023-03-11
No need to check current cell as it is 0

sak96:
# Pull Request Template

## Description

The cell value is 0 and only after the `check` function finishes.

## Type of change

- [x] Optimization (non-breaking change which alters functionality)

## Checklist:

- [ ] ~I ran `cargo clippy --all -- -D warning` just before my last commit and fixed any issue that was found.~ This is already done in #468


siriak (approved):
Looks good, thanks!

PR #470, opened 2023-03-11
Implement the A* search algorithm

qsantos:
## Description

This PR adds a generic implementation of the [A* search algorithm](https://en.wikipedia.org/wiki/A*_search_algorithm). This algorithm is very similar to Dijkstra's. The main difference is that it uses a heuristic to estimate the remaining distance (weight) from the current node to the target. As long as the heuristic is lower than the real value, the algorithm remains sound (returns the shortest path).

The existing implementation of Dijkstra's search algorithm does not return a path to a single target. Instead, it computes all shortest path from the given node. Since the A* algorithm requires a heuristic, it is hard to reason about the returned value when they are not a single node. For the sake of simplicity, this implementation takes a target node as input.

I have also made some minor changes to the implementation of Dijkstra's search algorithm, which just make the interface slightly nicer. I can remove the commit if you think it should be kept as-is.

## Type of change

## Checklist:

- [x] I ran `cargo clippy --all -- -D warnings` just before my last commit and fixed any issue that was found. (but see https://github.com/TheAlgorithms/Rust/pull/468)
- [x] I checked `COUNTRIBUTING.md` and my code follows its guidelines. ← **NOTE:** that should be `CONTRIBUTING.md`

The tests for this implementation of the A* search algorithm take about 150 ms together.

### DIFF
--- a/src/graph/astar.rs
+++ b/src/graph/astar.rs
@@ -1,0 +1,235 @@
+use std::{
+    collections::{BTreeMap, BinaryHeap},
+    ops::Add,
+};
+
+use num_traits::Zero;
+
+type Graph<V, E> = BTreeMap<V, BTreeMap<V, E>>;
+
+#[derive(Clone, Debug, Eq, PartialEq)]
+struct Candidate<V, E> {
+    estimated_weight: E,
+    real_weight: E,
+    state: V,
+}
+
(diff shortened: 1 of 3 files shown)


siriak (approved):
Looks good, thanks!

PR #475, opened 2023-03-19
added binary_to_decimal conversion

yosefahab:
# Pull Request Template

## Description

binary_to_decimal function, given a binary string of arbitrary length (ex: "00010"), convert it to a u128 decimal representation, the function may return `None` if an overflow is possible.
## Type of change

## Checklist:


siriak (approved):
Looks good, thanks!

PR #477, opened 2023-03-25
Add a payload to Union Find + some usage examples

aesteve:
# Add a payload to Union Find and some usage examples

## Description
I was reading through Union Find and was wondering how to implement it in Rust.
Thanks for this implementation, it was really helpful but I struggled a bit to understand how it'd work with anything else than a `usize`. It has been a bit confusing to me that we mix up the index in the array, and the actual data we may want to check disjoint sets / merge sets for. The example uses a contiguous set of integers, mixing the index in the array and the vaule itself. 

That's why I added a Map to bind the actual data we may want to use (Strings, etc.) with their index in the `parent_links` array.

On top of that, I added some Rust documentations 

## Type of change

- [X] Breaking change

I had to change the signature of some methods.

## Checklist:


siriak (approved):
Looks good, thanks!

PR #480, opened 2023-03-26
Added BLAKE2b hashing algorithm

dsmurrow:
# Pull Request Template

## Description

BLAKE2b is a performant cryptographic hashing algorithm tailored to 64-bit systems. This algorithm produces output digests between 1 and 64 bytes long.

## Type of change

## Checklist:


siriak (approved):
Looks good, thanks!

PR #481, opened 2023-03-29
Rework Topological sort

aesteve:
# Topological Sort, reworked

## Description

I was reading through Kahn's algorithm and topological sort and got a bit confused by the existing implementation, notably by the `visited` set which is declared, written to, but never read from.

So I went ahead and rewrote it the following way: 
* use a `Vec` of pairs to describe the graph, rather than a `BTreeMap` 
* relax the constraint for `Ord` for Nodes, and use Hash+Eq instead, using a `HashMap` instead of a `BTreeMap` (it sounds a bit more usual to me, but that's definitely opened to discussions)
* get rid of the `visited` set which wasn't used
* add a check at the end of the algorithm to check if we have visited every node,   
  * if so, return the topologically sorted nodes
  * if not, return an Error (=> means the algorithm now detects cycles, and doesn't only partially sort)
* I've added some tests, one coming from the Wikipedia page on topological sort, the other being a cyclic graph

Definitely open to discussions about the implementation and/or changes I've made. No problem, I mostly did this for learning purposes.

## Type of change

- [X] Breaking change since I changed the API a bit.

## Checklist:


siriak (approved):
Looks good, thanks!

PR #483, opened 2023-04-04
Rework Levenshtein distance

aesteve:
# Rework Levenshtein distance

## Description

Rework a bit the Levenshtein distance implementation by using a more Rust-idiomatic way (iterators).
Add a detailed documentation on how the algorithm works.

## Type of change

## Checklist:


aesteve:
Also, I think there's a duplicate between: 
* https://github.com/TheAlgorithms/Rust/blob/master/src/dynamic_programming/edit_distance.rs
* https://github.com/TheAlgorithms/Rust/blob/master/src/string/levenshtein_distance.rs

Not sure which one should take precedence but both are implementing Levenshtein distance.

src/string/levenshtein_distance.rs
```diff
@@ -1,42 +1,83 @@
...
-        let mut i = 1;
-        for c1 in string1.chars() {
-            let deletion_cost = d[i - 1] + 1;
-            let insertion_cost = d[i] + 1;
-            let substitution_cost = previous_substitution_cost + usize::from(c1 != c2);
+        for (col, c1) in string1.chars().enumerate() {
+            // col starts at 0
+            let deletion_cost = prev_dist[col] + 1; // "on the left" in the matrix (i.e. the value we just computed)
+            let insertion_cost = prev_dist[col + 1] + 1; // "on the top" in the matrix (means previous)
+            let substitution_cost = prev_substitution_cost + usize::from(c1 != c2); // top-left + 1 if char differs, 0 otherwise
```
siriak, line 61: Please use a condition instead, it'll be clearer

src/string/levenshtein_distance.rs
```diff
@@ -1,42 +1,83 @@
...
+        let mut prev_substitution_cost = prev_dist[0]; // we'll keep a reference to cell (i-1, j-1)
+        prev_dist[0] = row + 1; // diff with empty string, since `row` starts at 0, it's `row + 1`
 
-        let mut i = 1;
-        for c1 in string1.chars() {
-            let deletion_cost = d[i - 1] + 1;
-            let insertion_cost = d[i] + 1;
-            let substitution_cost = previous_substitution_cost + usize::from(c1 != c2);
+        for (col, c1) in string1.chars().enumerate() {
+            // col starts at 0
```
siriak, line 58: I imagine that's always the case in rust, so no need to mention it

src/string/levenshtein_distance.rs
```diff
@@ -1,42 +1,83 @@
...
+            let deletion_cost = prev_dist[col] + 1; // "on the left" in the matrix (i.e. the value we just computed)
+            let insertion_cost = prev_dist[col + 1] + 1; // "on the top" in the matrix (means previous)
+            let substitution_cost = prev_substitution_cost + usize::from(c1 != c2); // top-left + 1 if char differs, 0 otherwise
 
-            previous_substitution_cost = d[i];
-            d[i] = min3(deletion_cost, insertion_cost, substitution_cost);
+            prev_substitution_cost = prev_dist[col + 1]; // save the old value at (i-1, j-1)
 
-            i += 1;
+            // compute the minimum (uses a custom `min3` function for more readability)
```
siriak, line 65: No need for that comment, the code is self-explanatory

siriak (changes requested):
Please merge edit distance file into this one and delete the original. As for your changes, could you remove comments with example from the code? Examples don't need to be included if this can be looked up on Wikiipedia

aesteve:
Thanks for the feedback.
* I fixed the 3 reviewed items, thanks for that

As for the rest:
* I merged edit distance into the Levenshtein distance, but kept the entry in the DIRECTORY.md under Dynamic Programming to point at the Levenshtein distance:
  * in case some people are used to look for "edit distance" in the directory maybe? so that the change is a bit backwards compatible in a sense
  * and well, because Levenshtein distance is actually dynamic programming, indeed but works on Strings. So... It kinda makes sense.

I can remove it if you want, let me know. 

--- 

For the comments I removed them but I truly think this is a shame. Surely it can be looked be on Wikipedia, but they walked through the algorithm in a shorter yet understandable way.

siriak:
Thanks, looks good!

It's supposed to be as close to a real-life implementation as possible so that people can not only get familiar with the algorithm, but also with the language and coding conventions. So, it's different from a book and not supposed to be a complete self-explanatory text read as a book. We don't include details that are verbose and can be googled.

It's like when using patterns, saying that we use "visitor" pattern here is enough and an explanation of what that is can be easily googled. Hope that makes sense.

siriak (approved):

aesteve:
Makes total sense.

I probably need to rewrite the count-min-sketch ( #485 one) and a few others I was preparing (bloom filters, hyperloglog) as well then.

PR #484, opened 2023-04-04
Fixes clippy issues

aesteve:
# Pull Request Template

## Description

I noticed when running `cargo clippy --all --all-targets -- -D warnings` (adding `--all-targets` to include tests) that Clippy indicated some issues with the tests implementation.

I just applied Clippy's recommendation.

## Type of change

## Checklist:

- **[x] I ran `cargo clippy --all -- -D warnings` just before my last commit and fixed any issue that was found.**


siriak (approved):
Looks good, thanks!

PR #485, opened 2023-04-05
Count min sketch

aesteve:
# Count-Min Sketch

## Description

This is a first step towards adding more probabilistic data structures (like Bloom Filters for example).
This PR contains: 
* a new `probabilistic` module under the `data_structures` one
* a `CountMinSketch` trait containing the contract (basically, increment & return count)
* the default implementation using multiple hash functions
* **the introduction of `quickcheck` as dependency for Property-Based-Testing` which:
  * is really relevant in the case of probabilistic data structures because we can't test exact behavior but rather a property (in that case, a CountMinSketch must not underestimate the count, only overestimate it)
  * would (I think) benefit a lot in other modules like sorting. Since what we want to test in a lot of implementation is the "Oracle" property (for instance in all the `sort` implementations, we can just generate random inputs and compare with Rust's sort function. This could greatly simplify the tests. 

## Type of change

## Checklist:


DIRECTORY.md
```diff
@@ -47,6 +47,8 @@
     * [Treap](https://github.com/TheAlgorithms/Rust/blob/master/src/data_structures/treap.rs)
     * [Trie](https://github.com/TheAlgorithms/Rust/blob/master/src/data_structures/trie.rs)
     * [Union Find](https://github.com/TheAlgorithms/Rust/blob/master/src/data_structures/union_find.rs)
+      * Probabilistic Data Structures
+        * [Count-min Sketch](https://github.com/TheAlgorithms/Rust/blob/master/src/data_structures/probabilistic/count_min_sketch.rs)
```
siriak, line 51: Please correct indentation

siriak (commented):
The code looks good, thanks!

DIRECTORY.md
aesteve: Done.

siriak (approved):
Looks good, thanks!

PR #487, opened 2023-04-23
Permutation Algorithms

aesteve:
# Permutation Algorithms 

## Description

Adds a few common permutation algorithms: 
* the naive, recursive way, including a variant to avoid generating duplicates
* Heap's algorithm
* Steinhaus-Johnson-Trotter algorithm

## Type of change

## Checklist:

Note: tests are running under 300ms on my laptop, let me know if it's too slow on CI. If so, we can definitely reduce `MAX_SIZE` to 6 or less.


siriak (approved):
Looks good, thanks!

PR #501, opened 2023-05-25
Implemented Lazy Segment Tree

roger-sato:
## Description

The Lazy Segment Tree is a data structure used to perform range queries and updates on an array efficiently. It is a modification of the segment tree, which is a data structure that allows querying and updating elements in subarrays, or "segments," in O(log n) time, where n is the number of elements in the array. The Lazy Segment Tree introduces an optimization called "lazy propagation" to decrease the number of operations required for updates.

In the Lazy Segment Tree, instead of immediately updating all affected nodes of the tree during an update operation, the changes are "propagated" lazily. That is, the updates are stored at some node and only applied to its children when absolutely necessary (for example, during a query operation that affects one of those children). This allows update operations to be performed in O(log n) time as well, making the Lazy Segment Tree particularly suitable for scenarios where both queries and updates are frequent.

## Type of change

## Checklist:


siriak (approved):
Looks good, thanks!

PR #504, opened 2023-06-07
Add matrix chain multiplication

caojoshua:
# Pull Request Template

## Description

Well known problem to find the minimum number of multiplications to compute a chain of matrix multiplications. See the comments in code, or https://www.geeksforgeeks.org/matrix-chain-multiplication-dp-8/# for example. 

## Type of change

## Checklist:


siriak (approved):
Looks good, thanks!

PR #508, opened 2023-07-17
feat: add elementary geometry structs and algos

caojoshua:
# Pull Request Template

## Description

Add basic geometry algos / structs. I intend to use these to implement further geometry algos. For example, I've implemented Shamos-Hoey in https://github.com/caojoshua/Rust/commit/47a6955fdce72ae67939a377328fa101e0458a6a, which I would like to open as a separate PR if this can go through. I also see an existing `closest_points.rs`, which can probably be refactored to use the structs, but not in this PR.

## Type of change

## Checklist:

### DIFF
--- a/src/geometry/mod.rs
+++ b/src/geometry/mod.rs
@@ -1,3 +1,7 @@
 mod closest_points;
+mod point;
+mod segment;
 
 pub use self::closest_points::closest_points;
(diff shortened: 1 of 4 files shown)


siriak (approved):
Looks good, thanks!

PR #509, opened 2023-07-18
feat: add Shamos-Hoey algo

caojoshua:
# Pull Request Template

## Description

Shamos-Hoey algo that given a vector of segments, finds if any of the segments intersect.

I have some hacky code regarding how we compare keys for the btreeset, and I'm hoping I can get some rust help on how to better implement it. I'll drop a comment in the PR on this.

## Type of change

## Checklist:

### DIFF
--- a/src/geometry/shamos_hoey.rs
+++ b/src/geometry/shamos_hoey.rs
@@ -1,0 +1,309 @@
+/// Shamos-Hoey Algorithm. Given a set of line segments, finds if any two line segments intersect.
+/// Some implementations assume that there are no vertical lines. This implementation supports
+/// vertical lines.
+use super::point::Point;
+use super::segment::Segment;
+use std::cmp::{Ord, Ordering};
+use std::collections::BTreeSet;
+
+enum PointType {
+    Start,
+    End,
+}
+
+type SegmentAndPoint<'a> = (&'a Segment, &'a Point, PointType);
+struct SegmentWrapper<'a> {
+    segment: &'a Segment,
+    // `x` holds the x value when the segment was added to the set. We order segments at point `x`
+    // by their y-value at `x`. Note that if the order of two segments ever changes, it means that
+    // they intersected, which our algorithm can detect.
+    //
+    // Ideally we could track a single value of x as it changes through the algorithm iterations.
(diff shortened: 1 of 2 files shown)


src/geometry/shamos_hoey.rs
```diff
@@ -0,0 +1,309 @@
...
+            Ordering::Less | Ordering::Equal => (&self.a, &self.b),
+            Ordering::Greater => (&self.b, &self.a),
+        }
+    }
+}
+
+impl<'a> Ord for SegmentWrapper<'a> {
+    fn cmp(&self, other: &SegmentWrapper) -> Ordering {
+        // For vertical segments, we compare using the endpoint.
+        let x = if self.x < other.x { other.x } else { self.x };
```
caojoshua, line 52: In the algorithm, we should be using the latest value of x in the left-to-right sweep. However, the `Ord` implementation does not have access to the local variable x in `shamos_hoey()`,but we need to implement `Ord` if we want to use the stdlib btreeset implementation. The balanced tree implementations in this repo also compare elements using the `Ord` trait. ` I would strongly prefer to avoid using a ad-hoc balanced tree implementation. We get around this issue by storing x in the SegmentWrapper, (see the comment in the struct above).

I was hoping that I can implement `Ord` using a lambda that captures the local variable x, but I don't think there is a way of doing that in Rust. I also tried storing x into a global variable, but that produces inconsistent results among test threads.

Are there Rust experts who can help with making this code a little cleaner?

siriak: You might also want to write about it here https://discordapp.com/channels/808045925556682782/1022517736032714822. It's more likely to be discussed in Discord

caojoshua: Thank you. Dropped a post in https://discord.com/channels/808045925556682782/1022517736032714822/threads/1131472585390047232

caojoshua:
Not sure if there is still interest in getting this merged. I've asked for suggestions in discord, but there is no activity in the thread.

siriak:
That's unfortunate. You don't always get answers, such is life. Let's close this one, then, and feel free to open another one if you want to continue with that or something else. Thanks!

PR #510, opened 2023-07-23
feat: add graham scan algorithm to find convex hull

caojoshua:
# Pull Request Template

## Description

implement https://en.wikipedia.org/wiki/Graham_scan

## Type of change

## Checklist:


siriak (approved):
Looks good, thanks!

PR #515, opened 2023-07-28
number theory kth factor problem added

Irene-123:
## Description

Hii I have added a new category, that is, number theory and started off with Kth largest factor of number n. The program simply runs a for loop & output the factor. I've written a few tests, let me know if there are any changes.
[Reference](https://www.geeksforgeeks.org/kth-largest-factor-of-number-n/)

## Type of change

## Checklist:

There are no test cases currently which take long time


siriak (approved):
Looks good, thanks!

PR #518, opened 2023-08-12
Number theory

Irene-123:
# Pull Request Template

## Description

Added a new algorithm in Number Theory which is Euler's Totient Function for all numbers smaller than or equal to n. 
Referred from: [site](https://www.geeksforgeeks.org/eulers-totient-function-for-all-numbers-smaller-than-or-equal-to-n/)

Let me know if there are any issues. Thank you 

## Type of change

## Checklist:


siriak (approved):
Looks good, thanks!

PR #522, opened 2023-08-23
vignere algorithm 

mahmudsudo:
## Description

added a more expressive and idiomatic implementation of the algorithm


siriak (changes requested):
Could you put both implementations of vigenere at the same level? And have the same tests for them

PR #538, opened 2023-10-03
Added Intro Sort Algorithm

cyrixninja:
## Description

I implemented the Intro Sort (also known as Introspective Sort) Algorithm in Rust.Prior it was only available in Python and JavaScript. I tested the algorithm and got no warnings from the clippy. 

https://the-algorithms.com/algorithm/intro-sort
https://en.wikipedia.org/wiki/Introsort

## Type of change

## Checklist:


siriak (approved):
Looks good, thanks!

PR #540, opened 2023-10-03
Add sort utils fn and apply them to 3-way quick sort first.

suzp1984:
# Add sort utils methods

## Description

Add utils sort functions:

1. `generate_random_vec`: generate random `Vec<i32>` with size `n` and with elements within `[range_l, range_r]`;
2. `generate_nearly_ordered_vec`: generate nearly ordered `Vec<i32>` with size `n`, within range `[0, n-1], `swap_times` is to do how many times of swap on range `[0, n-1]`;
3. `generate_ordered_vec`: generate ordered `Vec<i32>` with size `n`, within range `[0, n-1]`;
4. `generate_reverse_ordered_vec`: generate reverse ordered `Vec<i32>` with size `n`, within range `[n-1, 0]`;
5. `generate_repeated_elements_vec`: generate `Vec<i32>` with size `n`, with repeated elements;
6. `log_timed`: log the `FnOnce`, the sort fn, execution time;

## Type of change

## Checklist:

## ~~Ignored Test `nearly_ordered_elements`~~
~~It takes around `10s` to run with array size `1000000`, there are some methods to improve the `quick sort`, e.g. do a random swap with the first element before doing every `partition`. I ignore this UT, and try to reopen it in another PR.~~

## improve 3-way quick sort for nearly ordered or pre-ordered array.
Solution: before doing partition in range [l, r], do random swap with `arr.swap(l, random(l, r))`.


src/sorting/quick_sort_3_ways.rs
```diff
@@ -44,60 +52,105 @@ mod tests {
...
+
         assert!(is_sorted(&res) && have_same_elements(&res, &cloned));
     }
 
     #[test]
     fn basic_string() {
         let mut res = vec!["a", "bb", "d", "cc"];
         let cloned = res.clone();
-        quick_sort_3_ways(&mut res);
+        sort_utils::log_timed("basic sring", || {
```
siriak, line 72:
```suggestion
        sort_utils::log_timed("basic string", || {
```

siriak (approved):
Looks good, thanks!

PR #548, opened 2023-10-07
Fix usize overflow 

suzp1984:
# Fix `usize` overflow problem when calculate middle of two `usize`

## Description

```
// left: usize;
// right: usize;
// left <= right;
let middle = (left + right) / 2;
```
if `(left + right) > std::usize::MAX`, then it's an overflow bug.

The common solution to calculate the middle value is:

```
let middle = left + (right - left) / 2;
```

## Type of change

## Checklist:


src/data_structures/lazy_segment_tree.rs
```diff
@@ -62,7 +62,7 @@ impl<T: Debug + Default + Ord + Copy + Display + AddAssign + Add<Output = T>> La
         if element_range.start >= query_range.start && element_range.end <= query_range.end {
             return Some(self.tree[idx]);
         }
-        let mid = (element_range.start + element_range.end) / 2;
+        let mid = element_range.start + (element_range.end - element_range.start) / 2;
```
siriak, line 65:
```suggestion
        let mid = element_range.start + (element_range.end - element_range.start) / 2; 
```
Just to restart PR checks

src/data_structures/lazy_segment_tree.rs
```diff
@@ -62,7 +62,7 @@ impl<T: Debug + Default + Ord + Copy + Display + AddAssign + Add<Output = T>> La
...
```
siriak, line 65:
```suggestion
        let mid = element_range.start + (element_range.end - element_range.start) / 2;
```

siriak (approved):
Looks good, thanks!

PR #549, opened 2023-10-07
Add a field trait and prime field implementation

boxdot:
## Description

The field trait will allow us to define an elliptic curve over any
field. We provide an implementation of the trait for a prime field of
order `P`, that is, the field isomorphic to the finite field `GF(P) =
ℤ/Pℤ`. This field will be used to define specific elliptic curves.

## Type of change

## Checklist:


siriak (approved):
Looks good, thanks!

PR #553, opened 2023-10-08
Count points on elliptic curve over finite prime field

boxdot:
## Description

Implemented via naive pre-calculation of all quadratic residues in the
prime field and via Legende symbol (the latter was conviniently already
implemented).

Also fix a typo and remove forgotten dbg! macros.

## Type of change

## Checklist:


siriak (approved):
Looks good, thanks!

PR #556, opened 2023-10-09
improve quick_sort index type to usize;

suzp1984:
# quick_sort improve: the index type should be `usize`.

## Description

the array's index should be `>=0`, so its type should be `usize` instead of `isize`.

## Type of change

## Checklist:


siriak (approved):
Looks good, thanks!

PR #560, opened 2023-10-10
Imporve heap

suzp1984:
# heap problem

## Description

* Let heap's data container vec start from the index zero, no need to waste one space.
* No need to let T impl Default trait if Vec starts from index zero;
* Iterator is not used to change the heap data, so I use `fn pop()` to replace it;
* Iterator for the min/max heap doesn't make sense, but I added an impl one anyway;
* smallest_child_idx fn don't make sense, it's the process of shift down, so I inline it inside the `fn pop()`;

I notice there are 

## Type of change

## Checklist:


ethanwater (commented):
✔️

PR #565, opened 2023-10-12
Implement Text Search Functionality in Rust

hardik-yadav1729:
This PR introduces a simple and efficient text search functionality in Rust. It allows users to search for a word within a given paragraph, offering the flexibility of performing either case-sensitive or case-insensitive searches.

 Please consider merging this pull request to make this functionality available to all users of the project.

Thank you for your time and consideration.


GreatRSingh:
There is no new line at the end of `mod.rs` file.

siriak:
This is not an implementation of a search algorithm (like KMP, Z-block, or naive), but rather an example of usage of `find` and we don't add such functions

Issue #559, opened 2023-10-09
Suggestion: Adding Loss functions of Machine Learning in maths folder


GreatRSingh:
I would like to suggest adding Loss functions in this repo.

The loss function estimates how well a particular algorithm models the provided data.

Add Loss Functions to: `machine_learning/loss_functions`

Task List:
- [x] Cross-Entropy
- [x] Hinge loss
- [x] Huber loss.
- [x] MSE
- [x] NLL
- [x] MAE
- [x] Marginal Ranking
- [x] KL Divergence

Navaneeth-Sharma:
Hi @GreatRSingh , 
Thanks for the suggestion. Just to give an update, Some of them already implemented in maths section. Please check that. I have taken this up to add as many functions as possible.

GreatRSingh:
@Navaneeth-Sharma Can you put up a list of loss functions that you are going to implement and which you have already implemented?

Navaneeth-Sharma:
Sure, 
Already implemented Algos
- ReLU
- Sigmoid
- Tanh

In Pull Request
- Leaky ReLU
- Softmax

In my List to implement
- ELU
- SELU

GreatRSingh:
@Navaneeth-Sharma ok I will do that.

Navaneeth-Sharma:
Hi, Just an info so that we dont implement the same losses, I will try to take up these loss functions 
- Cross-Entropy loss (bin class, multi class)
- Hinge loss
- Huber loss. 
Let me know if any of them are already taken @GreatRSingh

GreatRSingh:
@Navaneeth-Sharma No none of them are already taken.

I will take up MSE, NLL, MAE, MarginRanking, KLDivergence.

GreatRSingh:
@Navaneeth-Sharma @siriak I have added a list of Losses in the description.
Keep Suggesting any other if needed.

GreatRSingh:
Working on adding Hinge Loss

mobley-trent:
Are all these functions taken ?

GreatRSingh:
No, you can start working on them.

avats-dev:
Opened a PR to add KL divergence loss as a sub task of this issue. #656 
Kindly review.

TruongNhanNguyen:
I have opened a PR that implements Huber loss function, which is mentioned is this issue #697 :rocket:. 
Let's take a look on this PR :hugs:.

siriak:
[...]

PR #571, opened 2023-10-13
add `huber_loss` and `cross_entropy_loss` and doc for gelu function

Navaneeth-Sharma:
### Huber Loss Function:
Huber loss is a regression loss function that balances the benefits of Mean Squared Error (MSE) and Mean Absolute Error (MAE). It behaves like MSE for small differences between actual and predicted values and switches to MAE for larger differences, making it robust to outliers.

### Cross-Entropy Loss Function:
Cross-entropy loss is a classification loss function that measures the dissimilarity between actual and predicted probability distributions. It is widely used in tasks where data points need to be classified into different classes based on predicted probabilities.

covers part of #559 

Both functions contribute to the project's functionality, expanding its capabilities for machine learning and neural network applications.


siriak (approved):
Looks good, thanks!

PR #572, opened 2023-10-14
Added Baconian Cipher

cyrixninja:
## Description

Implemented Baconian Cipher which was prior only available in Python

References -
https://the-algorithms.com/algorithm/baconian-cipher
https://en.wikipedia.org/wiki/Bacon%27s_cipher

## Type of change

## Checklist:


siriak (approved):
Looks good, thanks!

PR #573, opened 2023-10-14
 Binomial Coefficient Calculation

hardik-yadav1729:
This pull request introduces an implementation of binomial coefficient calculation using BigInt for arbitrary precision arithmetic. The code also includes a suite of test cases, including three new random test cases, to ensure correctness and accuracy.

Please review!


siriak (approved):
Looks good, thanks!

PR #574, opened 2023-10-14
Implementation of Fizzy Number Program in Rust

hardik-yadav1729:
This pull request adds a Rust implementation of the Fizzy Number program, which calculates the n-th number that is a sum of powers of a given base. The code is well-documented with comments and includes five test cases to ensure correctness.

Hope adding this code in the repo would be helpful for the community.

Please review!


siriak (approved):
Looks good, thanks!

PR #582, opened 2023-10-18
Added mse_loss

GreatRSingh:
# Pull Request Template

## Description

Mean squared error (MSE) measures the amount of error in statistical models. It assesses the average squared difference between the observed and predicted values. When a model has no error, the MSE equals zero. As model error increases, its value increases. The mean squared error is also known as the mean squared deviation (MSD).

Covers: #559

## Type of change

## Checklist:


siriak (approved):
Looks good, thanks!

PR #598, opened 2023-11-05
tests: use `Poly1305::default` in `basic_tv1`

vil02:
# Pull Request Template

## Description
The [`Default` constructor of `Poly1305`](https://github.com/TheAlgorithms/Rust/blob/7ec4785a5ae6ce18bfb5b6eb993d7c222208108a/src/big_integer/poly1305.rs#L25) is not covered by any tests. This PR fixes that.

## Type of change
- [x] changed the existing test, to get _higher_ coverage.

## Checklist:

- [x] ~I added my algorithm to the corresponding `mod.rs` file within its own folder, and in any parent folder(s).~
- [x] ~I added my algorithm to `DIRECTORY.md` with the correct link.~


siriak (approved):
Looks good, thanks!

PR #599, opened 2023-11-06
tests: add test covering the `common_ratio == 1.0` case

vil02:
# Pull Request Template

## Description

This PR adds a test covering this line:
https://github.com/TheAlgorithms/Rust/blob/cef88b414f2d1b42551aa1bc5643f52be02feb37/src/math/sum_of_geometric_progression.rs#L8

I decided to update the structure of the corresponding tests and and express them like _parametrized_-like tests.

It seems that `graph::astar::tests::test_heuristic` [sometimes fails](https://github.com/TheAlgorithms/Rust/actions/runs/6776209984/job/18417199500#step:3:1030).

## Type of change

- [x] adds missing test case,

## Checklist:

- [x] ~I added my algorithm to the corresponding `mod.rs` file within its own folder, and in any parent folder(s).~
- [x] ~I added my algorithm to `DIRECTORY.md` with the correct link.~

### DIFF
--- a/src/math/sum_of_geometric_progression.rs
+++ b/src/math/sum_of_geometric_progression.rs
@@ -26,5 +26,8 @@ mod tests {
 
         let result_3 = sum_of_geometric_progression(9.0, 2.5, 5);
         assert_eq!(result_3, 579.9375);
+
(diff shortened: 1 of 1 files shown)


siriak (approved):
Looks good, thanks!

PR #602, opened 2023-11-08
Simplify bounds of `T` in `mode`

vil02:
# Pull Request Template

## Description

This PR simplifies the requirements on the type of the elements of the input of `mode` function. The added test case, shows that this is indeed a _generalization_.

It is related to #601.

## Type of change

## Checklist:


siriak (approved):
Looks good, thanks!

PR #603, opened 2023-11-09
Handle nonunique max frequency elements in `mode`

vil02:
# Pull Request Template

## Description

The current implementation of [`mode`](https://github.com/TheAlgorithms/Rust/blob/e7888f46dec0d96c764d710faa53994ac1d59e25/src/math/average.rs#L52) does not handle properly the inputs like `vec![1, 1, 2, 2]`.

This PR fixes that by changing the `mode` to return a `HashSet` of elements with maximal frequency.

Continuation of #602.

## Type of change

## Checklist:


siriak (approved):
Codecov works!

PR #612, opened 2023-11-13
Handle empty input in `median`

vil02:
# Pull Request Template

## Description

The [`median`](https://github.com/TheAlgorithms/Rust/blob/ed801737834477d35c39f35aa9eca20a2b062f1b/src/math/average.rs#L33C18-L33C18) function crashes for empty inputs. Since there is no `nan` for types integer types, I decided to return `Option<T>` from `median`. 

## Type of change

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected) (well, I change the return type, but _everything works_).

## Checklist:


siriak (approved):
Looks good, thanks!

PR #613, opened 2023-11-14
redundant condition removal

mapaba79:
Hi, this is my first PR in my life, I'm not a programmer, just a student and my English is very bad, but I noticed that this piece of code is redundant and I removed it. It's correct right?

### DIFF
--- a/src/math/amicable_numbers.rs
+++ b/src/math/amicable_numbers.rs
@@ -18,7 +18,7 @@ pub fn amicable_pairs_under_n(n: u32) -> Option<Vec<(u32, u32)>> {
     let mut out = vec![(0, 0)];
     // Check if numbers are amicable then append
     for (i, x) in factor_sums.iter().enumerate() {
-        if (*x != i as u32) && (*x < n) && (factor_sums[*x as usize] == i as u32) && (*x > i as u32)
+        if (*x < n) && (factor_sums[*x as usize] == i as u32) && (*x > i as u32)
         {
             out.push((i as u32, *x));
         }


src/math/amicable_numbers.rs
```diff
@@ -18,7 +18,7 @@ pub fn amicable_pairs_under_n(n: u32) -> Option<Vec<(u32, u32)>> {
     let mut out = vec![(0, 0)];
     // Check if numbers are amicable then append
     for (i, x) in factor_sums.iter().enumerate() {
-        if (*x != i as u32) && (*x < n) && (factor_sums[*x as usize] == i as u32) && (*x > i as u32)
+        if (*x < n) && (factor_sums[*x as usize] == i as u32) && (*x > i as u32)
         {
```
vil02, line 22: This should fix the formatting.

```suggestion
        if (*x < n) && (factor_sums[*x as usize] == i as u32) && (*x > i as u32) {
```

You can just _click_ the `commit suggestion` button.

vil02 (approved):
If `(*x > i as u32)` is satisfied the also `(*x != i as u32)`. So indeed it is redundant.

src/math/amicable_numbers.rs
mapaba79: I didn't even realize that this also fixed the formatting. Thanks for the feedback, I didn't know there was a commit suggestion button, just click on the line I want to comment on, right?

vil02: Now the formatting is _broken_. In order to fix it you need to remove the _newline symbol_ as suggested. You can do it in any way you want. One of them is described [here](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/reviewing-changes-in-pull-requests/incorporating-feedback-in-your-pull-request).

mapaba79: Ok I think I understand now, thanks again for your help. Now I think it's correct.

vil02 (approved):

PR #614, opened 2023-11-14
Handle empty input in `mode`

vil02:
# Pull Request Template

## Description

The current implementation of [`mode`](https://github.com/TheAlgorithms/Rust/blob/1f618c8067568e0c80e71d221ec3611b4f3be713/src/math/average.rs#L58) does not handle empty input. This PR fixes that.

I have been considering to return an empty `HashSet<T>` in case of an empty input, but I decided to change the return type to `Option<HashSet<T>>` and  return `None` in such situation.

## Checklist:


siriak (approved):
Looks good, thanks!

PR #634, opened 2023-12-13
LinkedList: make `get` sound

speedstyle:
`LinkedList::get` triggers an obvious use-after-free: it returns `&'static` references into nodes, which may later be dropped.

```rust
pub fn get(&mut self, index: i32) -> Option<&'static T> {
    Self::get_ith_node(self.head, index)
}
```

This PR modifies it to return `&T`, borrowing from `self` such that `delete` cannot be called until the reference is dead. I also changed it to take `&self` rather than `&mut self` so that multiple borrows can be taken out.

This is technically breaking, as uses which kept a reference while modifying the list will no longer compile. This would generally have been unsound though: reasonable uses (eg the unit tests) are unaffected.

## Type of change

## Checklist:

- [x] I ran below commands using the latest version of **rust nightly**.

### DIFF
--- a/src/data_structures/linked_list.rs
+++ b/src/data_structures/linked_list.rs
@@ -183,15 +183,15 @@ impl<T> LinkedList<T> {
         }
     }
 
-    pub fn get(&mut self, index: i32) -> Option<&'static T> {
-        Self::get_ith_node(self.head, index)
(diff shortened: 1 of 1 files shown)


siriak (approved):
Looks good, thanks!

PR #636, opened 2023-12-27
decrease Dijkstra upper bound from O(ElogE) to O(ElogV)

caojoshua:
# Pull Request Template

## Description

Before this patch, we put a new shortest path in a priority queue based on edges. This patch converts the priority queue to a btree with unique vertices mapped to its shortest path. When we find a new shortest path for a vertex, we either insert it into the tree, or update it. The tree's upper bound in space is O(V), allowing O(logV) pops and inserts.

## Type of change

## Checklist:

  - found some issues not related to my patch. could be mismatch in clippy versions.


siriak (approved):
Looks good, thanks!

PR #638, opened 2023-12-29
Added Simpson's rule and trapezoidal method for function integration

BogdanCiocea:
### Simpson's Rule
[Simpson's Rule](https://en.wikipedia.org/wiki/Simpson%27s_rule) is a numerical integration method used to approximate the definite integral of a function over a specified interval. The method is based on approximating the function as a series of quadratic polynomials and computing the area under these polynomials.

### Steps:
1. **Divide the interval [a, b] into n subintervals of equal width.**
````rust
h = (b - a) / n
````
2.For each subinterval, evaluate the function f at the left endpoint x0, right endpoint xn, and midpoint x1.
````rust
x0 = a + i * h
x1 = x0 + h / 2
x2 = b
````

3. **Calculate the Simpson's rule formula for each subinterval.**
````rust
integral = (h / 6) * (f(a) + 4 * f(x1) + 2 * f(x0) + 4 * f(x2) + f(b))
````

4. **Sum the Simpson's rule formula for each subinterval to obtain the estimated integral.**
````rust
total_integral = sum(integral)
````
### Advantages and Considerations:

- **Accuracy:**
  - Simpson's Rule generally provides more accurate results compared to simpler methods like the trapezoidal rule.

- **Convergence:**
  - The method converges relatively quickly as the number of subintervals (\(n\)) increases.

- **Even \(n\):**
  - For best results, \(n\) should be an even number.

- **Function Continuity:**
  - Simpson's Rule assumes the function is continuous and well-behaved over the interval.

### Example Usage:

```rust
let f = |x: f64| x.powi(2);
let a = 0.0;
let b = 1.0;
let n = 100;
let result = simpsons_rule(f, a, b, n);
```

## Trapezoidal Integral Implementation in Rust

This Rust code implements the trapezoidal rule for approximating the definite integral of a function f(x) between the limits x = a and x = b. The trapezoidal rule divides the interval between a and b into a specified number of trapezoids and approximates the integral as the sum of the areas of the trapezoids.

### Code Example

```rust
pub fn trapezoidal_integral<F>(a: f64, b: f64, f: F, precision: u32) -> f64
where
    F: Fn(f64) -> f64,
{
    let delta = (b - a) / precision as f64;

    let integral: f64 = (0..precision)
        .map(|trapezoid| {
            let left_side = a + (delta * trapezoid as f64);
            let right_side = left_side + delta;

            0.5 * (f(left_side) + f(right_side)) * delta
        })
        .sum();

    if a > b {
        -integral
    } else {
        integral
    }
}
```

### Description

The `trapezoidal_integral` function takes four arguments:

1. `a`: The lower bound of the integration interval

2. `b`: The upper bound of the integration interval

3. `f`: A function of the form `f(x)` that represents the integrand

4. `precision`: The number of trapezoids to use in the approximation

The function first defines a helper function `core` that performs the actual calculations for a single trapezoid. The `core` function takes the same four arguments as the `trapezoidal_integral` function and returns the area of a single trapezoid.

The `trapezoidal_integral` function then calculates the sum of the areas of the trapezoids using a for loop. The loop iterates over the range of `0..precision`, which represents the indices of the trapezoids. For each trapezoid, the `left_side` and `right_size` variables are calculated as the left and right endpoints of the trapezoid, respectively. The area of the trapezoid is then calculated using the formula:

````rust
0.5 * (f(left_side) + f(right_size)) * delta
````

where `delta` is the width of the trapezoid. The final step is to add the area of each trapezoid to the running sum and return the result.

### Usage

The `trapezoidal_integral` function can be used to approximate the definite integral of any function `f(x)` within a specified interval. For example, the following code approximates the integral of the function `x^2` from 0 to 1:

```rust
let f = |x: f64| x.powi(2);
let result = trapezoidal_integral(0.0, 1.0, f, 1000); // Approximates 1/3
println!("{}", result);
```

## Type of change

## Checklist:


src/sorting/multithreaded_bogo_sort.rs
```diff
@@ -0,0 +1,92 @@
...
+    // let elapsed_time = duration.recv().unwrap();
+    // println!("{:?}", elapsed_time);
+
+    // Join the handles in the vector
+    for handle in handles {
+        handle.join().unwrap();
+    }
+
+    // Sort the original data in-place using the last sorted result
+    data.sort();
```
siriak, line 51: What do you mean by `last sorted result` here? How do sorting threads mutate `data` array?

BogdanCiocea: The function creates multiple threads, each with a local copy of the data array. Threads repeatedly shuffle their local copies until the entire array is sorted or a global condition is met.

siriak: If they are all working with a local copy, how does the original array get sorted? We can just call `data.sort()` and don't bother with arrays if they don't change `data`

BogdanCiocea: Right. I'll modify it.

BogdanCiocea: Until then, can you please look into the simpson method?

siriak (approved):
Looks good, thanks!

PR #639, opened 2023-12-29
Least Square Approximation and Logarithm evaluation

mihaubuhai:
# Numerical Analysis Algorithms

## Description

Current pull request adds 2 new algorithms.

### Least Square Approximation
This algorithm is a response to the problem risen by analysis of some data, projected in grafing the data to points (in 2 dimension)
and figuring out a pattern that the data follow or fiding a median for those data.

Least Square Approximation creates a polynomial of any degree with the given set of points, starting from the following idea:
&ensp;&ensp;&ensp;&ensp;--> Fiding something in common for those points means, mathematically, to evaluate a distance; since we want to observe, relative to a polynomial, the behaviour of the data, we define a function ( called E ) that describes this idea:

&ensp;&ensp;&ensp;&ensp;&ensp;&ensp;&ensp;&ensp;&ensp;&ensp;&ensp;&ensp;&ensp;&ensp;&ensp;&ensp;&ensp;&ensp;&ensp;&ensp;&ensp;&ensp;&ensp;&ensp;&ensp;&ensp;&ensp;&ensp;&ensp;&ensp;&ensp;&ensp;&ensp;&ensp;&ensp;&ensp;&ensp;E(a<sub>0</sub>, a<sub>1</sub>, ... , a<sub>D</sub>) = $\sum_{i=0..n}$ (y<sub>i</sub> - P(x<sub>i</sub>))<sup>2</sup>, where:

&ensp;&ensp;&ensp;&ensp;&ensp;&ensp;P(..) -> polynomial we want to find; D -> degree of polynomial
&ensp;&ensp;&ensp;&ensp;&ensp;&ensp;n -> number of points (data)
&ensp;&ensp;&ensp;&ensp;&ensp;&ensp;y<sub>i</sub> -> ordinate of some data
&ensp;&ensp;&ensp;&ensp;&ensp;&ensp;x<sub>i</sub> -> abscissa of some data
&ensp;&ensp;&ensp;&ensp;&ensp;&ensp;a<sub>0</sub>, a<sub>1</sub>, ... , a<sub>D</sub> -> coefficients of terms of the polynomial

Now, for the idea to really produce a result, the function **E** shall be as small as possible, fact supported by smaller the sum, closer the values added, meaning y<sub>i</sub> and P(x<sub>i</sub>) are **very close to each other**. This now becomes a Calculus 2 problem, more specifically minimising the function **E**. This is possible by taking the partial derivative of the function "E" with respect to each coefficient **a<sub>i</sub>** and evaluate it to zero, as follows:

&ensp;&ensp;&ensp;&ensp;&ensp;&ensp;&ensp;&ensp;&ensp;&ensp;&ensp;&ensp;&ensp;&ensp;&ensp;&ensp;&ensp;&ensp;&ensp;&ensp;&ensp;&ensp;&ensp;&ensp;&ensp;&ensp;&ensp;&ensp;&ensp;&ensp;&ensp;&ensp;&ensp;&ensp;&ensp;&ensp;&ensp; $\frac{∂}{∂a_k}$ E(a<sub>0</sub>, a<sub>1</sub>, ..., a<sub>D</sub>) = 0 , for k = 0 .. D

Choosing a degree for **P** and replacing it's form into the formula for **E** above, we will have a system of equations, generated by the derivatives above, looking something like:

&ensp;&ensp;&ensp;&ensp;&ensp;&ensp;&ensp;&ensp;&ensp;&ensp;&ensp;&ensp;&ensp;&ensp;&ensp;&ensp;&ensp;&ensp;&ensp;&ensp;&ensp;&ensp;&ensp;&ensp;&ensp;&ensp;&ensp;&ensp;&ensp;&ensp;&ensp;&ensp;&ensp;&ensp;&ensp;&ensp;&ensp; $\sum_{i=0..D}$ **(a<sub>i</sub> * $\sum_{j=1..n}$ x<sub>j</sub> <sup>j+k</sup>)** = $\sum_{i=0..n}$ y<sub>i</sub>* x<sub>i</sub> <sup>k</sup> , for k = 0 .. D

Decomposing the bigger sum from the left side and exposing all the D + 1 equations, we can observe that the left side can be rewritten as follows:
$\to$ A matrix **A** which holds, as elements, $\sum_{j=1..n}$ x<sub>j</sub> <sup>j+k</sup> ( For each line, the elements will be the sums for the respective derivative (k representing it's index) )
$\to$ A column vector **x** which holds the coefficients $a_0, a_1, ..., a_D$, in this order 

Also, the right side can be expressed as a column vector **b**, which holds the sums $\sum_{i=0..n}$ y<sub>i</sub>* x<sub>i</sub> <sup>k</sup>

Finally, the whole algorithm is based on solving the equation: **A * x = b** for the column vector **x**; in this implementation, the method used is **QR** decomposition and ultimately solving it.

### Logarithm evaluation
This function calculates the log<sub>base</sub>x, for any base > 0, x > 0

For the case base = e and x between 0 and 1, the function uses the MacLaurin Series for ln: ln(x) = $\sum_{n>=1} {(-1)^n * \frac{x^n}{n}}$


siriak (approved):
Looks good, thanks!

PR #641, opened 2023-12-31
fix: use `is_nan`

vil02:
## Description

First of all: all the best for the new year (according to Gregorian calendar)!

This PR changes the tests of the [`log`](https://github.com/TheAlgorithms/Rust/blob/ebdc7cc5aef32f365d619024285a4c1448005660/src/math/logarithm.rs#L11) to use [`is_nan`](https://doc.rust-lang.org/std/primitive.f64.html#method.is_nan). This change makes the tests _more direct_ (i.e. we really want to check that `log` returns `nan` and not that rust _panics_ probably due to comparing two `nan`'s.).

@mihaubuhai - please have a look.

## Checklist:

### DIFF
--- a/src/math/logarithm.rs
+++ b/src/math/logarithm.rs
@@ -60,15 +60,13 @@ mod test {
     }
 
     #[test]
-    #[should_panic]
     fn test_log_zero_base() {
-        assert_eq!(log(0.0, 100.0, 0.00001), f64::NAN);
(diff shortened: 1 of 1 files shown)


mihaubuhai:
Yes :))) Didn't crossed my mind to think this way, thank you for looking into it!!!

siriak (approved):
Looks good, thanks!

PR #642, opened 2023-12-31
Added Bingo Sort

BogdanCiocea:
### Bingo Sort Implementation in Rust

The provided Rust code implements the bingo sort algorithm for sorting a vector of integers. The bingo sort algorithm repeatedly identifies and moves the "bingo" elements, which are elements that equal the current bingo value.

**Empty Vector Check**

- If the vector is empty, the function returns early to avoid further processing.

**Initializing Bingo Elements**

1. Two variables, `bingo` and `next_bingo`, are initialized with the first two elements of the vector.

**Identifying Maximum and Minimum**

1. The `max_min` function is called to identify the maximum and minimum elements in the vector.
2. The minimum element is stored in `next_bingo` for later use.

**Tracking Largest Element**

1. The `largest_element` variable is set to `next_bingo`, which is the current minimum element.

**Position of Next Element to Swap**

1. The `next_element_pos` variable keeps track of the position of the next element to swap with the current bingo element.

**Sorting Loop**

1. A loop iterates over the range of `bingo..=largest_element`, which represents the sequence of bingo elements.
2. Inside the loop:

   a. **Initializing Start Position:**

      - The `start_pos` variable is set to the current `next_element_pos` to serve as the starting position for iterating on the current bingo element.

   b. **Iterating on Remaining Elements**

      - The loop iterates over the remaining elements in the vector starting from the `start_pos` position.

         1. **Swapping Equal Elements:**

            - If the current element at index `i` equals the current bingo element `bingo`, then the elements at positions `i` and `next_element_pos` are swapped.

         2. **Updating Next Bingo Element:**

            - If the current element at index `i` is less than the current bingo element `bingo`, then it indicates that a new bingo element has been found.
            - The `next_bingo` variable is updated to the current element to prepare for the next iteration.

**Updating Bingo and Next Bingo Elements**

1. After the loop completes, the current and next bingo elements are updated:

   - The `bingo` variable is updated to the current `next_bingo` element.
   - The `next_bingo` variable is reset to the `largest_element` to start the next iteration with the next bingo element.

This iterative process continues until all bingo elements are sorted into their correct positions.

## Type of change

## Checklist:


siriak (approved):
Looks good, thanks!

PR #643, opened 2024-01-01
Trigonometric functions

mihaubuhai:
## Trig functions

This pull request provides implementations (generically) for the basic trig functions: **sine**, **cosine**, **tangent** and **cotangent**.

Those generic functions come with the condition for the argument (value for which the trig function will be evaluated) / that particular data type to implement **Into\<f64\>** and **Copy** traits.
Also my last 2 functions have been changed to generic types: [least_square_aprox](https://github.com/TheAlgorithms/Rust/blob/master/src/math/least_square_approx.rs) and [logarithm](https://github.com/TheAlgorithms/Rust/blob/master/src/math/logarithm.rs)

The trig functions **sine** and **cosine** are evaluated using their MacLaurin expansion, which is very much similar and that's the purpose of the 'template' function in trig_functions.rs file.
Because those trig functions are related to each other, their implementation is found in the same file, trig_functions.rs


siriak (approved):
Looks good, thanks!

PR #645, opened 2024-01-04
Properly handle flipped limits in `trapezoidal_integral`

vil02:
Current implementation of [`trapezoidal_integral`](https://github.com/TheAlgorithms/Rust/blob/88946160a3ee9a7770af607162ac8d69b93849a2/src/math/trapezoidal_integration.rs#L1) returns a wrong result, when integrating over an interval with _flipped_ limits. This PR:
- fixes that,
- adds a missing test case,
- rewrites tests as _parametrized tests_.

## Type of change

## Checklist:

### DIFF
--- a/src/math/trapezoidal_integration.rs
+++ b/src/math/trapezoidal_integration.rs
@@ -4,58 +4,39 @@ where
 {
     let delta = (b - a) / precision as f64;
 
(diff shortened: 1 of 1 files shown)


siriak (approved):
Looks good, thanks!

PR #650, opened 2024-01-08
Added Modular Exponential

tudor-cretu:
# Modular Exponential

### gcd_extended
The `gcd_extended` function implements the Extended Euclidean Algorithm, which is used to find the greatest common divisor (GCD) of two integers, as well as the coefficients of Bézout's identity. This is particularly useful in number theory and cryptography, especially for finding the modular inverse of an integer. Our implementation uses a recursive approach to calculate these values.

### mod_inverse
The `mod_inverse` function utilizes the `gcd_extended` function to compute the modular multiplicative inverse of an integer `b` modulo `m`. The modular inverse is an integer `x` such that `(b * x) % m = 1`. This function is essential for operations in modular arithmetic where division is required. The function will panic if the inverse does not exist, which occurs when `b` and `m` are not coprime.

### modular_exponential
The `modular_exponential` function performs modular exponentiation which is a type of exponentiation performed over a modulus. It is used to efficiently compute `base^power % modulus` for large integers. This is a fundamental operation in many cryptographic algorithms. Our function handles both positive and negative exponents, using the `mod_inverse` function for negative exponents to compute the result.

## Type of change

## Checklist:


siriak (approved):
Looks good, thanks!

PR #655, opened 2024-01-09
Cleanup `perfect_cube.rs`

vil02:
## Description
There are few problems with [`perfect_cube.rs`](https://github.com/TheAlgorithms/Rust/blob/cb3b21d81a0e97e4394cbb81db879d4e3f18f762/src/math/perfect_cube.rs):
- the [`perfect_cube`](https://github.com/TheAlgorithms/Rust/blob/cb3b21d81a0e97e4394cbb81db879d4e3f18f762/src/math/perfect_cube.rs#L1) is not tested (note the _typo_ in the [`test_perfect_cube`](https://github.com/TheAlgorithms/Rust/blob/cb3b21d81a0e97e4394cbb81db879d4e3f18f762/src/math/perfect_cube.rs#L43)),
- [`perfect_cube`](https://github.com/TheAlgorithms/Rust/blob/cb3b21d81a0e97e4394cbb81db879d4e3f18f762/src/math/perfect_cube.rs#L1) and [`perfect_cube_binary_search`](https://github.com/TheAlgorithms/Rust/blob/cb3b21d81a0e97e4394cbb81db879d4e3f18f762/src/math/perfect_cube.rs#L9) should have the same input type,
- [`perfect_cube`](https://github.com/TheAlgorithms/Rust/blob/cb3b21d81a0e97e4394cbb81db879d4e3f18f762/src/math/perfect_cube.rs#L1) does not give the correct result for `64 = 4*4*4`,
- I do not see any reason why to exclude negative numbers from being _perfect cubes_. For example `-8 = (-2)*(-2)*(-2)`, so why `-8` is not a _perfect cube_ [here](https://github.com/TheAlgorithms/Rust/blob/cb3b21d81a0e97e4394cbb81db879d4e3f18f762/src/math/perfect_cube.rs#L53)?

This PR fixes above issues by removing `perfect_cube` function and reorganizes the tests. 

## Type of change

## Checklist:

- [x] I ~added my~ removed algorithm ~to~ from the corresponding `mod.rs` file within its own folder, and in any parent folder(s).

### DIFF
--- a/src/math/perfect_cube.rs
+++ b/src/math/perfect_cube.rs
@@ -1,15 +1,8 @@
-pub fn perfect_cube(n: i32) -> bool {
-    // Calculate the cube root using floating-point arithmetic.
-    let val = (n as f64).powf(1.0 / 3.0);
-    // Check if the cube of the cube root equals the original number.
-    (val * val * val) == (n as f64)
-}
-
 // Check if a number is a perfect cube using binary search.
 pub fn perfect_cube_binary_search(n: i64) -> bool {
     // Handle negative numbers, as cube roots are not defined for negatives.
     if n < 0 {
-        return false;
+        return perfect_cube_binary_search(-n);
     }
 
     // Initialize left and right boundaries for binary search.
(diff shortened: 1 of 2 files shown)


src/math/perfect_cube.rs
```diff
@@ -1,15 +1,8 @@
...
-    // Check if the cube of the cube root equals the original number.
-    (val * val * val) == (n as f64)
-}
-
 // Check if a number is a perfect cube using binary search.
 pub fn perfect_cube_binary_search(n: i64) -> bool {
     // Handle negative numbers, as cube roots are not defined for negatives.
     if n < 0 {
-        return false;
+        return perfect_cube_binary_search(-n);
```
siriak, line 5: Shouldn't it be negative?

vil02: I am not sure what to you mean.

`n` is a perfect cube if and only if `-n` is a perfect cube. Based on that, the code is correct.

siriak: My bad, I thought it returned cubic root of the number

siriak (approved):

PR #657, opened 2024-01-10
Added Hexadecimal To Decimal Conversion

Adin3:
Added Hexadecimal to Decimal Conversion

the function takes a string slice that represents the hexadecimal number in string format and returns Result<u64, &`static str>

The function checks if the string slice is either empty or contains characters that cannot represent a hexadecimal number. In this case, the function will exit with Result::Err(error_message)

the function uses u64::from_str_radix() for the conversion of &str in base 16 to an u64.

If the function succeeds, it returns Result::Ok(decimal_number)

## Type of change

- Fixed one line in bingo_sort after a warning was caught by `cargo clippy --all -- -D warnings
- (checked afterward the unit tests, and everything was alright)
- Added Hexadecimal To Decimal Conversion

## Checklist:


siriak (approved):
Looks good, thanks!

PR #659, opened 2024-01-12
rgb_to_cmyk.rs and hsv_to_rgb.rs in conversions

tudor-cretu:
# RGB To CMYK and HSV To RGB

## Description

This implementation provides two essential color model conversion algorithms: RGB to CMYK and HSV to RGB. These algorithms are fundamental in image processing, graphics design, and computer vision, where it's often necessary to represent colors in different forms depending on the use case.

### 1) RGB To CMYK

The RGB (Red, Green, Blue) model is used primarily in electronic systems (like monitors and digital cameras) as it represents color as a combination of red, green, and blue light. On the other hand, the CMYK (Cyan, Magenta, Yellow, Black) model is used in color printing. It's a subtractive color model used in color printing where cyan, magenta, yellow, and black inks are applied in varying amounts to absorb light.

Conversion from RGB to CMYK involves calculating the percentage of each CMYK color needed to reproduce a color from the RGB model. The algorithm first converts RGB values (ranging from 0 to 255) to the range of 0 to 1 and then uses these normalized values to calculate the corresponding CMYK values.

### 2) HSV To RGB

HSV (Hue, Saturation, Value) is another color model that represents colors in terms of their hue, saturation, and brightness (value). It's often used in applications where it's more intuitive to separate color information (hue) from brightness and saturation.

The HSV to RGB conversion algorithm transforms colors from the HSV representation back to the RGB model. This conversion is particularly useful in situations where colors are modified or chosen based on their hue, saturation, and brightness, and then need to be converted back to the standard RGB model used for display purposes.

## Type of change

## Checklist:

### DIFF
--- a/src/conversions/hsv_to_rgb.rs
+++ b/src/conversions/hsv_to_rgb.rs
@@ -7,7 +7,10 @@
 // representation to the other.
 
 pub fn hsv_to_rgb(hue: f32, saturation: f32, value: f32) -> [i32; 3] {
-    if !(0.0..=360.0).contains(&hue) || !(0.0..=1.0).contains(&saturation) || !(0.0..=1.0).contains(&value) {
+    if !(0.0..=360.0).contains(&hue)
+        || !(0.0..=1.0).contains(&saturation)
+        || !(0.0..=1.0).contains(&value) 
+    {
         panic!("Input values out of range");
     }
 
--- a/src/conversions/hsv_to_rgb.rs
+++ b/src/conversions/hsv_to_rgb.rs
@@ -9,7 +9,7 @@
 pub fn hsv_to_rgb(hue: f32, saturation: f32, value: f32) -> [i32; 3] {
     if !(0.0..=360.0).contains(&hue)
         || !(0.0..=1.0).contains(&saturation)
-        || !(0.0..=1.0).contains(&value) 
+        || !(0.0..=1.0).contains(&value)
     {
         panic!("Input values out of range");
     }
@@ -19,7 +19,7 @@ pub fn hsv_to_rgb(hue: f32, saturation: f32, value: f32) -> [i32; 3] {
     let second_largest_component = chroma * (1.0 - (hue_section % 2.0 - 1.0).abs());
     let match_value = value - chroma;
 
-    let (red, green, blue) = if hue_section >= 0.0 && hue_section <= 1.0 {
+    let (red, green, blue) = if (0.0..=1.0).contains(&hue_section) {
         (chroma, second_largest_component, 0.0)
     } else if hue_section <= 2.0 {
         (second_largest_component, chroma, 0.0)
--- a/src/conversions/hsv_to_rgb.rs
+++ b/src/conversions/hsv_to_rgb.rs
@@ -1,0 +1,57 @@
+// The RGB color model is an additive color model in which red, green, and blue light
+// are added together in various ways to reproduce a broad array of colors. The name
+// of the model comes from the initials of the three additive primary colors, red,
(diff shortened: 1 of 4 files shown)


src/conversions/hsv_to_rgb.rs
```diff
@@ -0,0 +1,60 @@
...
+// light. In it, colors are represented using three components: hue, saturation and
+// (brightness-)value. This file provides functions for converting colors from one
+// representation to the other.
+
+pub fn hsv_to_rgb(hue: f32, saturation: f32, value: f32) -> [i32; 3] {
+    if !(0.0..=360.0).contains(&hue)
+        || !(0.0..=1.0).contains(&saturation)
+        || !(0.0..=1.0).contains(&value)
+    {
+        panic!("Input values out of range");
```
vil02, line 14: This line is not covered by tests.

src/conversions/hsv_to_rgb.rs
```diff
@@ -0,0 +1,60 @@
...
+        (second_largest_component, chroma, 0.0)
+    } else if hue_section <= 3.0 {
+        (0.0, chroma, second_largest_component)
+    } else if hue_section <= 4.0 {
+        (0.0, second_largest_component, chroma)
+    } else if hue_section <= 5.0 {
+        (second_largest_component, 0.0, chroma)
+    } else {
+        (chroma, 0.0, second_largest_component)
+    };
```
vil02, line 34: What do you think about using [`match`](https://doc.rust-lang.org/reference/expressions/match-expr.html) here?

src/conversions/hsv_to_rgb.rs
```diff
@@ -0,0 +1,60 @@
...
+
+    [
+        ((red + match_value) * 255.0).round() as i32,
+        ((green + match_value) * 255.0).round() as i32,
+        ((blue + match_value) * 255.0).round() as i32,
```
vil02, line 39: I would create a separate function doing the _mapping_ to `[0, 255]`.

src/conversions/rgb_to_cmyk.rs
```diff
@@ -0,0 +1,57 @@
...
+
+// Note: this is a very popular algorithm that converts colors linearly and gives
+// only approximate results. Actual preparation for printing requires advanced color
+// conversion considering the color profiles and parameters of the target device.
+
+pub fn rgb_to_cmyk(r_input: u8, g_input: u8, b_input: u8) -> (i32, i32, i32, i32) {
+    // Convert RGB values to the range 0..1
+    let r = r_input as f32 / 255.0;
+    let g = g_input as f32 / 255.0;
+    let b = b_input as f32 / 255.0;
```
vil02, line 12: Please define a separate function doing such _mapping_ to `[0.0, 1.0]`. It should also perform some _input check_.

src/conversions/rgb_to_cmyk.rs
```diff
@@ -0,0 +1,57 @@
...
+
+    let k = 1.0 - r.max(g).max(b);
+
+    if k == 1.0 {
+        return (0, 0, 0, 100); // pure black
+    }
+
+    let c = ((1.0 - r - k) / (1.0 - k) * 100.0).round() as i32;
+    let m = ((1.0 - g - k) / (1.0 - k) * 100.0).round() as i32;
+    let y = ((1.0 - b - k) / (1.0 - k) * 100.0).round() as i32;
```
vil02, line 22: What do you think about reducing the code duplication and having a separate function doing this calculations?

vil02 (changes requested):

vil02:
@tudor-cretu are you still working in this? If not, please close this PR.

PR #662, opened 2024-01-18
gaussian_quadrature

todi-mih:
This algorithm implements Gaussian Quadrature, a numerical integration technique, to approximate the definite integral of a given function over a specified interval. The gaussian_quadrature function takes the interval [a, b], a function f(x) to integrate, and the number of points to use in the quadrature.


todi-mih:
sorry about all the commits guys i was rushing

src/math/gaussian_quadrature.rs
```diff
@@ -0,0 +1,46 @@
+// Gaussian Quadrature Module
+//https://en.wikipedia.org/wiki/Gaussian_quadrature
+
+pub fn gaussian_quadrature(a: f64, b: f64, f: impl Fn(f64) -> f64, num_points: usize) -> f64 {
```
vil02, line 4:
```suggestion
pub fn gaussian_quadrature(a: f64, b: f64, f: impl Fn(f64) -> f64, order: usize) -> f64 {

```

src/math/gaussian_quadrature.rs
```diff
@@ -0,0 +1,46 @@
...
+
+pub fn gaussian_quadrature(a: f64, b: f64, f: impl Fn(f64) -> f64, num_points: usize) -> f64 {
+    let (points, weights) = get_gaussian_quadrature_points_weights(num_points);
+
+    let mut result = 0.0;
+
+    for i in 0..num_points {
+        let x_i = 0.5 * (a + b) + 0.5 * (b - a) * points[i];
+        result += weights[i] * f(x_i);
+    }
```
vil02, line 12:
```suggestion
    let result: f64 = points
        .iter()
        .zip(weights.iter())
        .map(|(&point, &weight)| {
            let x_i = 0.5 * ((a + b) + (b - a) * point);
            weight * f(x_i)
        })
        .sum();
```

src/math/gaussian_quadrature.rs
```diff
@@ -0,0 +1,46 @@
...
+    match num_points {
+        1 => (vec![0.0], vec![2.0]),
+        2 => (vec![-1.0 / 3.0, 1.0 / 3.0], vec![1.0, 1.0]),
+        3 => (
+            vec![-0.7745966692414834, 0.0, 0.7745966692414834],
+            vec![0.5555555555555556, 0.8888888888888888, 0.5555555555555556],
+        ),
+        _ => unimplemented!("Gaussian Quadrature not implemented for this number of points"),
+    }
+}
```
vil02, line 28: I would highly suggest to generalize it for any _number of points_/_order_. It also seems that it uses Legendre polynomials. It also could be more general.

src/math/gaussian_quadrature.rs
```diff
@@ -0,0 +1,46 @@
...
+        let a = 0.0;
+        let b = 1.0;
+        let f = |x: f64| x.powi(2);
+        let num_points = 3;
+        let expected = 1.0 / 3.0;
+        let eps = 0.0001;
+
+        let result = gaussian_quadrature(a, b, f, num_points);
+        assert!((result - expected).abs() < eps);
+    }
```
vil02, line 45: What do you think about testing with _flipped_ integration limits?

```suggestion

        let result_neg = gaussian_quadrature(b, a, f, num_points);
        assert!((result_neg + expected).abs() < eps);
    }
```

vil02 (changes requested):

src/math/gaussian_quadrature.rs
todi-mih: Nice  suggestion,the test passes will add it as well

src/math/gaussian_quadrature.rs
todi-mih: Hmm,i am kinda cofused here ,can you please tell me more information about what i should add?

vil02: Let's start with basic stuff: it would be great to call `gaussian_quadrature` with `order` being `20` and obtain a valid result.

todi-mih: Ahh, i see what you mean,sorry for being late ,been kinda busy,thanks for the explanation my man.

todi-mih: i keep trying to add the options but the result just keeps getting more complicated and wrong ,my purpose in here wat to give an example of another numerical approach to integrals,after some research 3 is the most command order for this method (and most accurate) ,mind if i leave it like this?

vil02: I think we should go for the general case.

vil02:
@todi-mih are you still working on this? If not, please close this PR.

PR #663, opened 2024-01-18
Huffman alg

todi-mih:
The algorithm implements Huffman coding, a data compression technique. It encodes a given text into a binary representation using variable-length codes for each character. The encoded text and the mapping of characters to their Huffman codes are returned. The algorithm also provides a function to decode the encoded text back to the original form, ensuring lossless compression


DIRECTORY.md
```diff
@@ -184,6 +184,7 @@
     * [Frizzy Number](https://github.com/TheAlgorithms/Rust/blob/master/src/math/frizzy_number.rs)
     * [Gaussian Elimination](https://github.com/TheAlgorithms/Rust/blob/master/src/math/gaussian_elimination.rs)
     * [Gaussian Error Linear Unit](https://github.com/TheAlgorithms/Rust/blob/master/src/math/gaussian_error_linear_unit.rs)
+    * [Gaussian_quadrature](https://github.com/Rust/blob/new-feature-branch/src/math/gaussian_quadrature.rs)
```
vil02, line 187:
```suggestion
```

src/compression/huffman.rs
```diff
@@ -0,0 +1,114 @@
...
+    },
+}
+
+pub fn huffman_encode(text: &str) -> (String, BTreeMap<char, String>) {
+    let mut frequencies = HashMap::new();
+
+    for char in text.chars() {
+        let counter = frequencies.entry(char).or_insert(0);
+        *counter += 1;
+    }
```
vil02, line 22:
```suggestion
    let frequencies = text.chars().fold(HashMap::new(), |mut map, val| {
        *map.entry(val).or_insert(0) += 1;
        map
    });
```
I would suggest to move it into a separate function.

src/compression/huffman.rs
```diff
@@ -0,0 +1,114 @@
...
+
+    let mut heap = BinaryHeap::new();
+    for (char, freq) in frequencies.iter() {
+        heap.push(HuffmanNode::Leaf {
+            character: *char,
+            frequency: *freq,
+        });
+    }
```
vil02, line 30:
```suggestion
    let mut heap: BinaryHeap<_> = frequencies
        .iter()
        .map(|(&char, &freq)| HuffmanNode::Leaf {
            character: char,
            frequency: freq,
        })
        .collect();
```

As above: I would suggest a separate function.

src/compression/huffman.rs
```diff
@@ -0,0 +1,114 @@
...
+                    HuffmanNode::Internal { frequency: f2, .. },
+                )
+                | (
+                    HuffmanNode::Leaf { frequency: f1, .. },
+                    HuffmanNode::Leaf { frequency: f2, .. },
+                ) => f1 + f2,
+            },
+            left: Box::new(left),
+            right: Box::new(right),
+        };
```
vil02, line 56:
```suggestion
        let internal_node = HuffmanNode::Internal {
            frequency: left.frequency() + right.frequency(),
            left: Box::new(left),
            right: Box::new(right),
        };
```

Make sure to add:
```rust
impl HuffmanNode {
    fn frequency(&self) -> usize {
        match *self {
            HuffmanNode::Internal { frequency, .. } | HuffmanNode::Leaf { frequency, .. } => {
                frequency
            }
        }
    }
}
```

vil02 (changes requested):

PR #669, opened 2024-01-21
Redo derivative

Github-Vektor:
# Pull Request Template

## Description
The derivative method, in calculus, calculates how a function's output changes as its input varies. It's represented as f'(x) or dy/dx and is used to find slopes, rates of change, and solve optimization problems

## Type of change

## Checklist:


src/sorting/bogo_bogo_sort.rs
alexandruradovici, line 1: Where is this file used?

PR #672, opened 2024-02-06
Assign value directly for newly created node in `BinarySearchTree::insert`.

iamlockon:
## Description
Not sure if it's a too tiny pick. Change the tree node value assignment to be inline.

## Type of change

## Checklist:


vil02 (approved):
Nicely spotted!

PR #677, opened 2024-02-14
update `selection_sort`

genskyff:
# Pull Request Template

## Description

- Return immediately if the length is less than `2`
- Only need to iterate up to `len - 1` instead of `len`
- Only perform the swap when `left != smallest`

## Checklist:

### DIFF
--- a/src/sorting/selection_sort.rs
+++ b/src/sorting/selection_sort.rs
@@ -1,13 +1,22 @@
 pub fn selection_sort<T: Ord>(arr: &mut [T]) {
     let len = arr.len();
-    for left in 0..len {
+
+    if len < 2 {
+        return;
+    }
+
+    for left in 0..(len - 1) {
         let mut smallest = left;
+
         for right in (left + 1)..len {
             if arr[right] < arr[smallest] {
                 smallest = right;
             }
         }
-        arr.swap(smallest, left);
+
+        if left != smallest {
+            arr.swap(left, smallest);
+        }
     }
 }
 
--- a/src/sorting/selection_sort.rs
+++ b/src/sorting/selection_sort.rs
@@ -1,22 +1,13 @@
 pub fn selection_sort<T: Ord>(arr: &mut [T]) {
     let len = arr.len();
-
-    if len < 2 {
-        return;
-    }
-
     for left in 0..(len - 1) {
         let mut smallest = left;
-
         for right in (left + 1)..len {
             if arr[right] < arr[smallest] {
                 smallest = right;
             }
         }
-
-        if left != smallest {
-            arr.swap(left, smallest);
-        }
+        arr.swap(smallest, left);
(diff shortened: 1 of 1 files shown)


src/sorting/selection_sort.rs
```diff
@@ -1,13 +1,22 @@
 pub fn selection_sort<T: Ord>(arr: &mut [T]) {
     let len = arr.len();
-    for left in 0..len {
+
+    if len < 2 {
+        return;
+    }
```
vil02, line 6: This only makes the code more complex. Note that if `len  < 2`, then we will not enter the loop.

src/sorting/selection_sort.rs
```diff
@@ -1,13 +1,22 @@
...
         for right in (left + 1)..len {
             if arr[right] < arr[smallest] {
                 smallest = right;
             }
         }
-        arr.swap(smallest, left);
+
+        if left != smallest {
+            arr.swap(left, smallest);
+        }
```
vil02, line 19: This `if` seems to be _inside_  the `swap` method. cf. [docs](https://doc.rust-lang.org/std/primitive.slice.html#method.swap).

src/sorting/selection_sort.rs
```diff
@@ -1,13 +1,22 @@
 pub fn selection_sort<T: Ord>(arr: &mut [T]) {
     let len = arr.len();
-    for left in 0..len {
+
+    if len < 2 {
+        return;
+    }
+
+    for left in 0..(len - 1) {
```
vil02, line 8: I making this loop _one shorter_ makes sense.

vil02 (changes requested):
I do not see a real point of this change. Please have a look at my comments below.

genskyff:
Thanks for the feedback! I'll make the suggested changes and update the PR soon.

src/sorting/selection_sort.rs
```diff
@@ -1,6 +1,10 @@
 pub fn selection_sort<T: Ord>(arr: &mut [T]) {
+    if arr.is_empty() {
+        return;
+    }
+
     let len = arr.len();
-    for left in 0..len {
+    for left in 0..(len - 1) {
```
genskyff, line 7: Although there is some additional code here, I believe it makes it easier for people to understand. The textbook's description of selection sort states that the first loop only needs to go up to `len - 1`. I think this modification will not cause confusion.

vil02: The original code is simply shorter and handles all of the cases. I really do not see a reason why to introduce a new _branch_.

genskyff: @vil02 Thank you for your feedback. I agree that the original code is more succinct and covers all cases. Therefore, I have decided to close this PR. Thank you for your time and advice!

PR #680, opened 2024-03-07
Single Layer Perceptron

943n:
## Description

This pull request adds a Perceptron implementation according to the provided template. The Perceptron is a simple supervised learning algorithm used for binary classification tasks. It consists of a single layer neural network with adjustable weights that are updated during training to minimize classification errors.

The implementation includes methods for initializing the Perceptron, training it with provided input-output pairs, and testing its accuracy. The Perceptron uses a predefined activation function `ActivationFunction::None` indicating no activation function is applied during its operation.

## Type of change

## Checklist:

- [x] I checked `COUNTRIBUTING.md` and my code follows its guidelines. (If you haven't checked, please do so)

### DIFF
--- a/src/machine_learning/perceptron/activation_function.rs
+++ b/src/machine_learning/perceptron/activation_function.rs
@@ -1,0 +1,23 @@
+pub enum ActivationFunction {
+    Sigmoid,
+    Step,
+    Tanh,
+    None,
+}
+
+impl ActivationFunction {
+    pub fn activate(&self, x: f64) -> f64 {
+        match self {
+            ActivationFunction::Sigmoid => 1.0 / (1.0 - x.exp()),
+            ActivationFunction::Step => {
+                if x > 0.0 {
+                    1.0
+                } else {
+                    -1.0
+                }
+            }
+            ActivationFunction::Tanh => x.tanh(),
+            ActivationFunction::None => x,
+        }
+    }
+}
--- a/src/machine_learning/perceptron/activation_function.rs
+++ b/src/machine_learning/perceptron/activation_function.rs
@@ -1,25 +1,25 @@
 // Enum representing different activation functions for the perceptron model.
 pub enum ActivationFunction {
-    Sigmoid,  // Sigmoid activation function
-    Step,     // Step (or Heaviside) activation function
-    Tanh,     // Hyperbolic tangent activation function
-    None,     // No activation function (identity)
+    Sigmoid, // Sigmoid activation function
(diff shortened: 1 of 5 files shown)


src/machine_learning/perceptron/single_layer_perceptron.rs
```diff
@@ -0,0 +1,103 @@
...
+
+pub struct Perceptron {
+    weights: Vec<f64>,                 // Weights assigned to each input feature
+    learning_rate: f64,                // Learning rate for weight adjustment during training
+    activation_fn: ActivationFunction, // Activation function used to compute the output
+}
+
+impl Perceptron {
+    // Initialize a new perceptron with random weights, learning rate, and activation function
+    pub fn new(input_size: usize, learning_rate: f64, activation_fn: ActivationFunction) -> Self {
```
vil02, line 16: What do you think about having such signature:
```suggestion
    pub fn new(input_size: usize, learning_rate: f64, activation_fn: F) -> Self 
    where
         F: Fn(f64) -> f64,
    {
```
Instead of the name `F` you can use `ActivationFunction`.
There is not need to have the class `ActivationFunction`.

src/machine_learning/perceptron/activation_function.rs
vil02, line 1: I am not sure if this is needed. I would suggest to use some _function pointer_ while creating the `Perceptron` object.

src/machine_learning/perceptron/single_layer_perceptron.rs
```diff
@@ -0,0 +1,103 @@
+// Perceptron model implementation.
+// A perceptron is a basic unit of a neural network, capable of binary classification.
+// It takes inputs, applies weights, and passes the result through an activation function to produce an output.
+
+use crate::machine_learning::perceptron::ActivationFunction;
+use rand::prelude::*;
+
+pub struct Perceptron {
+    weights: Vec<f64>,                 // Weights assigned to each input feature
+    learning_rate: f64,                // Learning rate for weight adjustment during training
```
vil02, line 10: Why do you have this field here? A `Perceptron` should not know about its `learning_rate`. This is used only while learning. Teaching and using a `Perceptron` are two different things and should not be mixed.

vil02 (changes requested):
I am not sure if this class is suitable for this repo - not much can be done with a single perceptron.

The design of the class `Perceptron` needs some improvement. Currently it covers _training_ and _using_ and this should be split.

PR #684, opened 2024-03-14
added a decremental connectivity data structure

Miraksi:
# Decremental Connectivity Query in Rust

## Description

A data structure that, given a forest, allows dynamic-connectivity queries. Meaning deletion of an edge (u,v) and checking whether two vertecies are still connected. This implementation of an algorithm developed by Shimon Even and Yossi Shiloach, has a precomputiation phase with complexity O(n) (n being the number of verticies), a deletion complexity of O(log n) and a connectivity-query complexity of O(1).

## New feature

## Checklist:


src/graph/decremental_connectivity.rs
```diff
@@ -0,0 +1,179 @@
...
+
+        if self.is_smaller(u, v) {
+            queue.push(u);
+            self.dfs_id += 1;
+            self.visited[v] = self.dfs_id;
+        } else {
+            queue.push(v);
+            self.dfs_id += 1;
+            self.visited[u] = self.dfs_id;
+            println!("entered");
```
vil02, line 56: Is this needed?
```suggestion
```

src/graph/decremental_connectivity.rs
```diff
@@ -0,0 +1,179 @@
...
+        assert_eq!(dec_con1.connected(5, 0), Some(true));
+        assert_eq!(dec_con1.connected(2, 7), Some(false));
+        assert_eq!(dec_con1.connected(0, 9), None);
+        dec_con1.delete(0, 2);
+        assert_eq!(dec_con1.connected(3, 4), Some(true));
+        assert_eq!(dec_con1.connected(5, 0), Some(false));
+        assert_eq!(dec_con1.connected(5, 6), Some(true));
+        assert_eq!(dec_con1.connected(8, 7), Some(true));
+        dec_con1.delete(7, 8);
+        assert_eq!(dec_con1.connected(8, 7), Some(false));
```
vil02, line 169:
```suggestion
        assert!(dec_con1.connected(3, 4).unwrap());
        assert!(dec_con1.connected(5, 0).unwrap());
        assert!(!dec_con1.connected(2, 7).unwrap());
        assert!(dec_con1.connected(0, 9).is_none());
        dec_con1.delete(0, 2);
        assert!(dec_con1.connected(3, 4).unwrap());
        assert!(!dec_con1.connected(5, 0).unwrap());
        assert!(dec_con1.connected(5, 6).unwrap());
        assert!(dec_con1.connected(8, 7).unwrap());
        dec_con1.delete(7, 8);
        assert!(!dec_con1.connected(8, 7).unwrap());
```

src/graph/decremental_connectivity.rs
```diff
@@ -0,0 +1,179 @@
...
+        dec_con1.delete(7, 8);
+        dec_con1.delete(1, 4);
+
+        let mut dec_con2 = super::DecrementalConnectivity::new(&adjacent);
+        dec_con2.delete(4, 1);
+
+        let mut dec_con3 = super::DecrementalConnectivity::new(&adjacent);
+        dec_con3.delete(1, 4);
```
vil02, line 177: Are here some assertions missing?

src/graph/decremental_connectivity.rs
```diff
@@ -0,0 +1,179 @@
...
+
+    pub fn connected(&self, u: usize, v: usize) -> Option<bool> {
+        match (self.component.get(u), self.component.get(v)) {
+            (Some(a), Some(b)) => Some(a == b),
+            _ => None,
+        }
+    }
+
+    // original adjacency will not be modified by this function
+    // expects for the graph to have an edge (u,v)
```
vil02, line 40: What do you think about using `panic!` to ensure that?

vil02 (changes requested):

src/graph/decremental_connectivity.rs
Miraksi: oops, that was an oversight

src/graph/decremental_connectivity.rs
Miraksi: I'd love that, but detecting whether there is an edge between u and v would be in O(n) since, we do not have an adjacency matrix, but rather for each vertex we store the vertices it is adjacent to.

Still looking for a fix

Miraksi: In the case where u and v are not immediately adjacent, `delete(u,v)` will still disconnect both vertices in O(log n), by deleting either an edge connected to u, or v. This means, that this function would even then be of value, so maybe just renaming the function to `disconnect(u,v)` would also be an option. 
The only problem i see with this is the ambiguity of the deletion. We can only be certain, that an edge will be deleted and not which edge, but we give the power to the developer to decide, what they want.
What do you think?

src/graph/decremental_connectivity.rs
Miraksi: Done

vil02:
@Miraksi could you please share some reliable source about this data structure?

Miraksi:
> @Miraksi could you please share some reliable source about this data structure?

grafik
Shimon Even and Yossi Shiloach,
An On-Line Edge Deletion Problem,
SIAM Journal on Computing, 1972

This also is the paper took as reference, which was also referenced in the wikipedia entry on decremental connectivity

vil02:
@Miraksi your implementation works only for forests (_connected list of trees_), but the it is possible to (mis)use it for any graph. Do you see some reasonable way to avoid that?

By the way: feel free to _rebase_.

Miraksi:
> @Miraksi your implementation works only for forests (_connected list of trees_), but the it is possible to (mis)use it for any graph. Do you see some reasonable way to avoid that?
> 
> By the way: feel free to _rebase_.

Good Idea! I've added a function that checks whether the given graph is acyclic (ignoring cycles of lenght one). This check runs in O(n) time, so its basically for free.
Should the constructor panic, or return a result type?

src/graph/decremental_connectivity.rs
vil02: I am not sure if `disconnect(u,v)` is a good idea. I am simply not sure what what will be the resulting object after such operation.

What do you think about storing all of the edges in some hashset? Correct me if I am wrong, but this will not change the time and space complexity.

Miraksi: deletion will now panic, if the edge doesn't exist :)
I've also switched to HashSets, as you recommended

src/graph/decremental_connectivity.rs
```diff
@@ -0,0 +1,228 @@
...
+
+    pub fn connected(&self, u: usize, v: usize) -> Option<bool> {
+        match (self.component.get(u), self.component.get(v)) {
+            (Some(a), Some(b)) => Some(a == b),
+            _ => None,
+        }
+    }
+
+    pub fn delete(&mut self, u: usize, v: usize) {
+        println!("{:?}", self.component);
```
vil02, line 45:
```suggestion
```

src/graph/decremental_connectivity.rs
```diff
@@ -0,0 +1,228 @@
...
+/// Meaning deletion of an edge (u,v) and checking whether two vertecies are still connected.
+///
+/// # Complexity
+/// The preprocessing phase runs in O(n) time, where n is the the number of vertecies in the forest.
+/// Deletion runs in O(log n) and checking for connectivity runs in O(1) time.
+///
+/// # Sources
+/// used Wikipedia as reference: <https://en.wikipedia.org/wiki/Dynamic_connectivity>
+pub struct DecrementalConnectivity {
+    adjacent: Vec<HashSet<usize>>,
```
vil02, line 13: I think you are assuming that your graph is not directed, i.e. each edge is in two directions. Is there a check somewhere that if `self.adjacent[a].contains(&b)`, then `self.adjacent[b].contains(&a)`?

src/graph/decremental_connectivity.rs
```diff
@@ -0,0 +1,228 @@
...
+        }
+    }
+
+    pub fn delete(&mut self, u: usize, v: usize) {
+        println!("{:?}", self.component);
+        if !self.adjacent[u].contains(&v) || self.component[u] != self.component[v] {
+            panic!(
+                "delete called on the edge ({}, {}) which doesn't exist",
+                u, v
+            );
```
vil02, line 50: This is not _covered_ by any test.

vil02 (changes requested):

src/graph/decremental_connectivity.rs
Miraksi: added a check in `has_cycle()`

vil02 (approved):
Looks good enough, thanks!

PR #685, opened 2024-03-16
Handle unicode characters in `duval_algorithm`

realstealthninja:
Fix duval_algorithm.rs on unicode strings being entered

## Description
I've replaced all instances of &[u8] with its character equivalent and replaced the other instances 
with appropriate substitutes

## Type of change

## Checklist:

> note:
> cargo clippy does produce errors but none of them were part of the file I was working on. I'd appreciate it if someone took at look at it


vil02 (approved):
I will push one more commit to _reduce code duplication_ in the tests of `duval_algorithm`.

Issue #683, opened 2024-03-14
String based algorithms panics on unicode characters


realstealthninja:
Many of the string algorithms where str.as_bytes() is used are prone to this error 
so far i have tested and the issue is present with 
- [x] ~duval_algorithm~ done in #685 
- [x] ~knuth_morris_pratt~ done in #686
- [x] rabin_karp
- [x] z_algoritm

realstealthninja:
duval was fixed in the pr #685

PR #686, opened 2024-03-17
Handle unicode characters in `knuth_morris_pratt`

realstealthninja:
# handle unicode in knuth morris pratt as well as reduce code duplication

## Description
Part of #683 
I've replaced instances of `.as_bytes` with `.chars` to handle unicode characters.
I have also taken the liberty from the last pull request to reduce the code duplication present in the tests
using a macro.

## Type of change
## Checklist:

### DIFF
--- a/src/string/knuth_morris_pratt.rs
+++ b/src/string/knuth_morris_pratt.rs
@@ -40,63 +40,29 @@ pub fn knuth_morris_pratt(st: &str, pat: &str) -> Vec<usize> {
 mod tests {
     use super::*;
 
-    #[test]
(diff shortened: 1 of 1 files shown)


vil02 (approved):
I allowed myself to push two small changes:
- 262f134b5abfba868021b8d5cfa36f4c963aee28 - please note that is `as_bytes` would be use, the test would fail,
- 4dca7b0674658600a17f53ef9bcc474c6fdb4c18 - `index` is not needed.

Side remark: try not to use the `master`/`main` branch for tour changes.

Thanks for the fix.

PR #687, opened 2024-03-17
feat: Add test to increase coverage of naive permutations

realstealthninja:
#  Add a test to cover the empty array case in the naive implementation of permutations

## Description
I've opted for adding a test unsure if this is the right way to go. Do advice for better alternatives
## Type of change

- [ ] ~New feature (non-breaking change which adds functionality)~
- [ ] ~Breaking change (fix or feature that would cause existing functionality to not work as expected)~

## Checklist:

### DIFF
--- a/src/general/permutations/naive.rs
+++ b/src/general/permutations/naive.rs
@@ -99,6 +99,7 @@ mod tests {
     #[test]
     fn empty_array() {
         let empty: std::vec::Vec<u8> = vec![];
+        assert_eq!(permute(&empty), vec![vec![]]);
         assert_eq!(permute_unique(&empty), vec![vec![]]);
     }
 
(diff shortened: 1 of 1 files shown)


src/general/permutations/naive.rs
```diff
@@ -96,6 +96,12 @@ mod tests {
         }
     }
 
+    #[test]
+    fn empty_array() {
+        let empty: std::vec::Vec<u8> = vec![];
+        assert_eq!(permute_unique(&empty), vec![vec![]]);
```
vil02, line 102: What do you think about explicitly testing `permute` against an empty input?

```suggestion
        assert_eq!(permute(&empty), vec![vec![]]);
        assert_eq!(permute_unique(&empty), vec![vec![]]);
```

vil02 (changes requested):

vil02 (approved):
Thanks!

PR #689, opened 2024-03-20
KMP solves the shortest palindrome

mengfansheng-git:
KMP solves the shortest palindrome


src/string/shortest_palindrome.rs
vil02, line 1: Please add some documentation, so that potential user or reader will know what is this function doing.

src/string/shortest_palindrome.rs
```diff
@@ -0,0 +1,48 @@
+pub fn shortest_palindrome(s: String) -> String {
```
vil02, line 1: Why not like that?
```suggestion
pub fn shortest_palindrome(s: &str) -> String {
```

src/string/shortest_palindrome.rs
```diff
@@ -0,0 +1,48 @@
...
+    if s.is_empty() {
+        return "String is Empty!".to_string();
```
vil02, line 3: An empty string `""` is a palindrome. I think in this case an empty string should be returned:
```suggestion
        return "".to_string();
```

src/string/shortest_palindrome.rs
```diff
@@ -0,0 +1,48 @@
...
+    }
+
+    #[test]
+    fn example_two() {
+        assert_eq!(shortest_palindrome("abcd".to_string()), "dcbabcd");
+    }
+    #[test]
+    fn example_three() {
+        assert_eq!(shortest_palindrome("".to_string()), "String is Empty!");
+    }
```
vil02, line 47: The duplicated code in the tests could be reduced. What do you think about the following?

```suggestion
    use super::shortest_palindrome;
    use crate::string::is_palindrome;

    macro_rules! test_shortest_palindrome {
        ($($name:ident: $inputs:expr,)*) => {
            $(
                #[test]
                fn $name() {
                    let (s, expected) = $inputs;
                    assert!(is_palindrome(expected));
                    assert_eq!(shortest_palindrome(s), expected);
                    assert_eq!(shortest_palindrome(expected), expected);
                }
            )*
        }
    }

    test_shortest_palindrome! {
        empty: ("", ""),
        extend_left_1: ("aacecaaa", "aaacecaaa"),
        extend_left_3: ("abcd", "dcbabcd"),
        // unicode_0: ("അ", "അ"),
    }
```

vil02 (changes requested):
Please make sure that the Unicode characters are handled correctly (#683). There is some explanation needed: for the input: `"abcd"` why the output is `"dcbabcd"` and not `"abcdcba"`?

src/string/shortest_palindrome.rs
mengfansheng-git: Based mainly on https://oi-wiki.org/string/kmp/

src/string/shortest_palindrome.rs
```diff
@@ -0,0 +1,54 @@
...
+                assert_eq!(shortest_palindrome(expected), expected);
+            }
+        )*
+    }
+}
+test_shortest_palindrome! {
+    empty: ("", ""),
+    extend_left_1: ("aacecaaa", "aaacecaaa"),
+    extend_left_3: ("abcd", "dcbabcd"),
+}
```
vil02, line 54:
```suggestion
#[cfg(test)]
mod tests {
    use super::shortest_palindrome;
    macro_rules! test_shortest_palindrome {
        ($($name:ident: $inputs:expr,)*) => {
            $(
                #[test]
                fn $name() {
                    use crate::string::is_palindrome;
                    let (s, expected) = $inputs;
                    assert!(is_palindrome(expected));
                    assert_eq!(shortest_palindrome(s), expected);
                    assert_eq!(shortest_palindrome(expected), expected);
                }
            )*
        }
    }
    test_shortest_palindrome! {
        empty: ("", ""),
        extend_left_1: ("aacecaaa", "aaacecaaa"),
        extend_left_3: ("abcd", "dcbabcd"),
    }
}
```

src/string/shortest_palindrome.rs
```diff
@@ -0,0 +1,54 @@
...
+                assert_eq!(shortest_palindrome(s), expected);
+                assert_eq!(shortest_palindrome(expected), expected);
+            }
+        )*
+    }
+}
+test_shortest_palindrome! {
+    empty: ("", ""),
+    extend_left_1: ("aacecaaa", "aaacecaaa"),
+    extend_left_3: ("abcd", "dcbabcd"),
```
vil02, line 53: Please add tests with Unicode characters.

vil02 (changes requested):

src/string/shortest_palindrome.rs
```diff
@@ -0,0 +1,57 @@
+/*
+The function shortest_palindrome expands the given string to shortest palindrome by adding a shortest prefix.
+KMP. Source：https://www.scaler.com/topics/data-structures/kmp-algorithm/
+Prefix Functions and KPM. Source：https://oi-wiki.org/string/kmp/
+*/
+
+pub fn shortest_palindrome(s: &str) -> String {
+    if s.is_empty() || !s.is_ascii() {
+        return s.to_string();
+    }
```
vil02, line 10: I see that you are using TDD :smile:.

src/string/shortest_palindrome.rs
```diff
@@ -0,0 +1,57 @@
...
+            }
+        )*
+    }
+}
+test_shortest_palindrome! {
+    empty: ("", ""),
+    extend_left_1: ("aacecaaa", "aaacecaaa"),
+    extend_left_2: ("abcd", "dcbabcd"),
+    unicode_1: ("അ", "അ"),
+    unicode_2: ("牛", "牛"),
```
vil02, line 56: I think one should add at least one more test case:
```suggestion
    unicode_2: ("牛", "牛"),
    unicode_3: ("a牛", "牛a牛"),
```

vil02 (changes requested):

src/string/shortest_palindrome.rs
```diff
@@ -0,0 +1,57 @@
...
+
+    let p_chars: Vec<char> = s.chars().collect();
+    let mut suffix = vec![0; s.chars().count()];
+    for i in 1..s.chars().count() {
+        let mut j = suffix[i - 1];
+        while j > 0 && p_chars[j] != p_chars[i] {
+            j = suffix[j - 1];
+        }
+        suffix[i] = j + if p_chars[j] == p_chars[i] { 1 } else { 0 };
+    }
```
vil02, line 20: What happens here? I think that it could be moved to a separate function.

src/string/shortest_palindrome.rs
```diff
@@ -0,0 +1,57 @@
...
+    let mut dp = vec![0; s.chars().count()];
+    let mut s_chars: Vec<char> = s.chars().rev().collect();
+    dp[0] = if p_chars[0] == s_chars[0] { 1 } else { 0 };
+    for i in 1..s_chars.len() {
+        let mut j = dp[i - 1];
+        while j > 0 && s_chars[i] != p_chars[j] {
+            j = suffix[j - 1];
+        }
+        dp[i] = j + if s_chars[i] == p_chars[j] { 1 } else { 0 };
+    }
```
vil02, line 31: Same as above.

src/string/shortest_palindrome.rs
```diff
@@ -0,0 +1,57 @@
...
+
+    s_chars.append(&mut p_chars[dp[s.chars().count() - 1]..s.chars().count()].to_vec());
+    s_chars.iter().collect()
```
vil02, line 34: Same as above.

vil02 (changes requested):

src/string/shortest_palindrome.rs
mengfansheng-git: Iterating through "abcd" like this via s.len() is 4，but  "a牛" like this via s.len() is 4，Can create transgressions，Here you need to iterate through the number of string characters，So it's changed to "s.chars().count()"

vil02: Sure.

A very good way to explain the reader what happens in a piece of code is to _move_ it into a new function with proper name.

src/string/shortest_palindrome.rs
```diff
@@ -0,0 +1,73 @@
...
+    dp[0] = if p_chars[0] == s_chars[0] { 1 } else { 0 };
+    for i in 1..s.chars().count() {
+        let mut j = dp[i - 1];
+        while j > 0 && s_chars[i] != p_chars[j] {
+            j = suffix[j - 1];
+        }
+        dp[i] = j + if s_chars[i] == p_chars[j] { 1 } else { 0 };
+    }
+    dp
+}
```
vil02, line 51: There are few changes in this _monster suggestion_:

```suggestion
pub fn shortest_palindrome(s: &str) -> String {
    if s.is_empty() {
        return "".to_string();
    }

    let p_chars: Vec<char> = s.chars().collect();
    let suffix = raw_suffix_function(&p_chars);

    let mut s_chars: Vec<char> = s.chars().rev().collect();
    // The prefix of the original string matches the suffix of the flipped string.
    let dp = invert_suffix_function(&p_chars, &s_chars, &suffix);

    s_chars.append(&mut p_chars[dp[p_chars.len() - 1]..p_chars.len()].to_vec());
    s_chars.iter().collect()
}

pub fn raw_suffix_function(p_chars: &Vec<char>) -> Vec<usize> {
    let mut suffix = vec![0; p_chars.len()];
    for i in 1..p_chars.len() {
        let mut j = suffix[i - 1];
        while j > 0 && p_chars[j] != p_chars[i] {
            j = suffix[j - 1];
        }
        suffix[i] = j + if p_chars[j] == p_chars[i] { 1 } else { 0 };
    }
    suffix
}

pub fn invert_suffix_function(p_chars: &[char], s_chars: &[char], suffix: &[usize]) -> Vec<usize> {
    let mut dp = vec![0; p_chars.len()];
    dp[0] = if p_chars[0] == s_chars[0] { 1 } else { 0 };
    for i in 1..p_chars.len() {
        let mut j = dp[i - 1];
        while j > 0 && s_chars[i] != p_chars[j] {
            j = suffix[j - 1];
        }
        dp[i] = j + if s_chars[i] == p_chars[j] { 1 } else { 0 };
    }
    dp
}
```

src/string/shortest_palindrome.rs
```diff
@@ -0,0 +1,73 @@
...
+    if s.is_empty() {
+        return "".to_string();
+    }
+
+    let p_chars: Vec<char> = s.chars().collect();
+    let suffix = raw_suffix_function(s, p_chars.clone());
+
+    let mut s_chars: Vec<char> = s.chars().rev().collect();
+    // The prefix of the original string matches the suffix of the flipped string.
+    let dp = invert_suffix_function(s, p_chars.clone(), s_chars.clone(), suffix);
```
vil02, line 17: One should avoid clones. Here cloning is not needed.

src/string/shortest_palindrome.rs
```diff
@@ -0,0 +1,73 @@
...
+
+    s_chars.append(&mut p_chars[dp[s.chars().count() - 1]..s.chars().count()].to_vec());
```
vil02, line 19: computing `s.chars().count()` is very expensive. The number of characters can be obtained from `p_chars`.

vil02 (changes requested):

src/string/shortest_palindrome.rs
```diff
@@ -0,0 +1,73 @@
...
+        )*
+    }
+}
+test_shortest_palindrome! {
+    empty: ("", ""),
+    extend_left_1: ("aacecaaa", "aaacecaaa"),
+    extend_left_2: ("abcd", "dcbabcd"),
+    unicode_1: ("അ", "അ"),
+    unicode_2: ("a牛", "牛a牛"),
+}
```
vil02, line 73:
```suggestion
#[cfg(test)]
mod tests {
    use crate::string::shortest_palindrome;
    macro_rules! test_shortest_palindrome {
    ($($name:ident: $inputs:expr,)*) => {
        $(
            #[test]
            fn $name() {
                use crate::string::is_palindrome;
                let (s, expected) = $inputs;
                assert!(is_palindrome(expected));
                assert_eq!(shortest_palindrome(s), expected);
                assert_eq!(shortest_palindrome(expected), expected);
            }
        )*
        }
    }
    test_shortest_palindrome! {
        empty: ("", ""),
        extend_left_1: ("aacecaaa", "aaacecaaa"),
        extend_left_2: ("abcd", "dcbabcd"),
        unicode_1: ("അ", "അ"),
        unicode_2: ("a牛", "牛a牛"),
    }
}
```

mengfansheng-git: Thank you. I've benefited from your advice.

PR #698, opened 2024-03-25
Implement Subset Sum algorithm using Dynamic Programming.

TruongNhanNguyen:
## Description

In this PR, I use dynamic programming approach to implement Subset Sum problem.

## Type of change

## Checklist:

### DIFF
--- a/src/dynamic_programming/subset_sum.rs
+++ b/src/dynamic_programming/subset_sum.rs
@@ -6,7 +6,7 @@
 ///
 /// * `nums` - A slice of integers representing the input array.
 /// * `target_sum` - The target sum that the subset needs to achieve.
-pub fn subset_sum(nums: &[i32], target_sum: i32) -> (bool, Option<Vec<i32>>) {
+pub fn subset_sum(nums: &[u32], target_sum: u32) -> (bool, Option<Vec<u32>>) {
     let n = nums.len();
     let mut dp = vec![vec![false; (target_sum + 1) as usize]; n + 1];
 
@@ -19,7 +19,7 @@ pub fn subset_sum(nums: &[i32], target_sum: i32) -> (bool, Option<Vec<i32>>) {
     for (i, num) in nums.iter().enumerate() {
         for j in 1..=target_sum as usize {
             dp[i + 1][j] = dp[i][j];
-            if *num <= j as i32 {
+            if *num <= j as u32 {
                 dp[i + 1][j] |= dp[i][j - *num as usize];
             }
         }
--- a/src/dynamic_programming/subset_sum.rs
+++ b/src/dynamic_programming/subset_sum.rs
@@ -1,49 +1,109 @@
-/// Determines whether there exists a subset of the given array `nums` that adds up to the specified `target_sum`.
-/// If such a subset exists, returns the subset that adds up to `target_sum`, sorted in ascending order.
-/// If no such subset exists, returns `None`.
+/// Computes subsets of a given array of numbers that sum up to a target sum.
 ///
 /// # Arguments
 ///
-/// * `nums` - A slice of integers representing the input array.
-/// * `target_sum` - The target sum that the subset needs to achieve.
-pub fn subset_sum(nums: &[u32], target_sum: u32) -> Option<Vec<u32>> {
-    let n = nums.len();
-    let mut dp = vec![vec![false; (target_sum + 1) as usize]; n + 1];
+/// * `nums` - A slice of unsigned 32-bit integers representing the input numbers.
+/// * `target_sum` - The target sum to which subsets are sought.
+///
+/// # Returns
+///
+/// An option containing a vector of vectors representing subsets that sum up to the target sum,
+/// or `None` if no such subsets exist.
+pub fn subset_sum(nums: &[u32], target_sum: u32) -> Option<Vec<Vec<u32>>> {
+    let mut dp = initialize_dp_table(nums.len(), target_sum);
 
-    // Base case: if the target sum is 0, then it's always possible to achieve it with an empty subset.
+    fill_dp_table(nums, &mut dp);
+
+    extract_subsets(&dp, nums.len(), target_sum)
+}
+
+/// Initializes the DP table with empty vectors.
+///
+/// # Arguments
+///
+/// * `rows` - The number of rows in the DP table.
+/// * `columns` - The number of columns in the DP table.
+///
+/// # Returns
+///
+/// A vector of vectors of empty vectors representing the DP table.
+fn initialize_dp_table(rows: usize, columns: u32) -> Vec<Vec<Vec<Vec<u32>>>> {
+    // Initialize DP table with empty vectors
+    let mut dp = vec![vec![vec![]; (columns + 1) as usize]; rows + 1];
+
+    // Base case: Empty subset is a valid solution for target sum 0
     for row in &mut dp {
-        row[0] = true;
+        row[0] = vec![vec![]];
     }
 
-    // Fill the dp table
+    dp
+}
+
+/// Populates the DP table with subsets that sum up to specific values.
+///
+/// # Arguments
+///
+/// * `nums` - A slice of unsigned 32-bit integers representing the input numbers.
+/// * `dp` - A mutable reference to the DP table.
+fn fill_dp_table(nums: &[u32], dp: &mut [Vec<Vec<Vec<u32>>>]) {
     for (i, num) in nums.iter().enumerate() {
-        for j in 1..=target_sum as usize {
-            dp[i + 1][j] = dp[i][j];
-            if *num <= j as u32 {
-                dp[i + 1][j] |= dp[i][j - *num as usize];
-            }
+        for j in 1..=dp[0].len() - 1 {
+            dp[i + 1][j] = update_subsets(dp, i, *num, j as u32);
+        }
+    }
+}
+
+/// Updates subsets in the DP table for a given number and target sum.
+///
+/// # Arguments
+///
+/// * `dp` - A reference to the DP table.
+/// * `i` - Index representing the current number being considered.
+/// * `num` - The current number being considered.
+/// * `j` - The target sum for which subsets are sought.
+///
+/// # Returns
+///
+/// A vector of vectors representing subsets that sum up to the target sum.
+fn update_subsets(dp: &[Vec<Vec<Vec<u32>>>], i: usize, num: u32, j: u32) -> Vec<Vec<u32>> {
+    let mut new_subsets = dp[i][j as usize].clone();
+
+    // If the current number is less than or equal to the target sum,
+    // update subsets with the current number
+    if num <= j {
+        let prev_subsets = dp[i][(j - num) as usize].clone();
+        for prev_subset in prev_subsets {
+            let mut new_subset = prev_subset.clone();
+            new_subset.push(num);
+            new_subsets.push(new_subset);
         }
     }
 
-    // Reconstruct the subset
-    let mut subset = Vec::new();
-    let mut i = n;
-    let mut j = target_sum as usize;
-    while i > 0 && j > 0 {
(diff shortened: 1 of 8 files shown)


src/dynamic_programming/subset_sum.rs
```diff
@@ -0,0 +1,89 @@
...
+        (false, None)
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_subset_sum_exists() {
```
realstealthninja, line 55: since nums is i32 could you add a test for negative numbers as well it negative numbers arent covered please use the unsigned equivalent

realstealthninja (changes requested):

src/dynamic_programming/subset_sum.rs
TruongNhanNguyen: Well, I forgot that this algorithm requires non-negative integers as input. I have updated the function signature that use unsigned integer type `u32` in Rust.

Miraksi:
Is there a reason for your function to return `(bool, Option<Vec<u32>>)` and not just `Option<u32>`?
Since you only ever return `(true, Some(x))` and `(false, None)` I see no point in also returning a bool.

TruongNhanNguyen:
> Is there a reason for your function to return `(bool, Option<Vec<u32>>)` and not just `Option<u32>`? Since you only ever return `(true, Some(x))` and `(false, None)` I see no point in also returning a bool.

The function will return a `bool` value to specify that exist subset (or non-exist). If bool value is true, we can return subset of the array that has sum of elements equal to given `target`, else return `None`.

realstealthninja (approved):
LGTM @vil02  could you have a look please.

src/dynamic_programming/subset_sum.rs
```diff
@@ -0,0 +1,85 @@
...
+
+    // Fill the dp table
+    for (i, num) in nums.iter().enumerate() {
+        for j in 1..=target_sum as usize {
+            dp[i + 1][j] = dp[i][j];
+            if *num <= j as u32 {
+                dp[i + 1][j] |= dp[i][j - *num as usize];
+            }
+        }
+    }
```
vil02, line 26: Please move it into a separate function.

src/dynamic_programming/subset_sum.rs
```diff
@@ -0,0 +1,85 @@
...
+    let mut subset = Vec::new();
+    let mut i = n;
+    let mut j = target_sum as usize;
+    while i > 0 && j > 0 {
+        if dp[i][j] && !dp[i - 1][j] {
+            subset.push(nums[i - 1]);
+            j -= nums[i - 1] as usize;
+        }
+        i -= 1;
+    }
```
vil02, line 38: Please move is into a separate function.

src/dynamic_programming/subset_sum.rs
```diff
@@ -0,0 +1,85 @@
...
+        assert_eq!(subset, None);
+    }
+
+    #[test]
+    fn test_subset_sum_single_element_array() {
+        let nums = vec![7];
+        let target_sum = 7;
+        let subset = subset_sum(&nums, target_sum);
+        assert_eq!(subset, Some(vec![7]));
+    }
```
vil02, line 84:
```suggestion
    macro_rules! test_subset_sum {
        ($($name:ident: $inputs:expr,)*) => {
            $(
                #[test]
                fn $name() {
                    let (nums, target_sum, expected) = $inputs;
                    assert_eq!(subset_sum(&nums, target_sum), expected);
                }
            )*
        }
    }

    test_subset_sum! {
        subset_sum_exists: (vec![3, 34, 4, 12, 5, 2], 9, Some(vec![4, 5])),
        sum_does_not_exist: (vec![3, 34, 4, 12, 5, 2], 30, None),
        empty_array: (vec![], 5, None),
        length_1_array: (vec![7], 7, Some(vec![7])),
        length_1_array_no_solution: (vec![7], 8, None),
    }
```

vil02 (changes requested):

src/dynamic_programming/subset_sum.rs
```diff
@@ -0,0 +1,85 @@
+/// Determines whether there exists a subset of the given array `nums` that adds up to the specified `target_sum`.
+/// If such a subset exists, returns the subset that adds up to `target_sum`, sorted in ascending order.
+/// If no such subset exists, returns `None`.
+///
+/// # Arguments
+///
+/// * `nums` - A slice of integers representing the input array.
+/// * `target_sum` - The target sum that the subset needs to achieve.
+pub fn subset_sum(nums: &[u32], target_sum: u32) -> Option<Vec<u32>> {
```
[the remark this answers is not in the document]
TruongNhanNguyen: - To identifies and handles duplicate elements correctly, when updating subsets in the `dp` table, we ensures that each subset is unique by using a clone operation to create new subsets. This ensures that duplicate elements do not affect the generation of subsets.
- Non-unique solutions refer to different combinations of elements that yield the same sum. We can generate all possible subsets that sum up to the target sum, including those with duplicate elements, and then removes duplicate subsets before returning the result. This ensures that the result contains only distinct subsets, each representing a unique combination of elements.

src/dynamic_programming/subset_sum.rs
```diff
@@ -0,0 +1,85 @@
...
+    let n = nums.len();
```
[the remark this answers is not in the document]
TruongNhanNguyen: I have refactored the code base, so this unneeded `n` in no longer used.

src/dynamic_programming/subset_sum.rs
TruongNhanNguyen: Done

src/dynamic_programming/subset_sum.rs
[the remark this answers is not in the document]
TruongNhanNguyen: You can see updates on the latest commit.

src/dynamic_programming/subset_sum.rs
TruongNhanNguyen: Done.

src/dynamic_programming/subset_sum.rs
```diff
@@ -0,0 +1,134 @@
...
+/// * `target_sum` - The target sum to which subsets are sought.
+///
+/// # Returns
+///
+/// An option containing a vector of vectors representing subsets that sum up to the target sum,
+/// or `None` if no such subsets exist.
+pub fn subset_sum(nums: &[u32], target_sum: u32) -> Option<Vec<Vec<u32>>> {
+    let mut dp = initialize_dp_table(nums.len(), target_sum);
+
+    fill_dp_table(nums, &mut dp);
```
vil02, line 15: I think should look somehow like this:
```suggestion
    let dp = create_dp(nums, target_sum);
```
Furthermore, `dp` is not really a good name.

src/dynamic_programming/subset_sum.rs
```diff
@@ -0,0 +1,134 @@
...
+        }
+    }
+
+    test_subset_sum! {
+        test_subset_sum_basic: (vec![1, 2, 3, 4, 5], 5, Some(vec![vec![1, 4], vec![2, 3], vec![5]])),
+        test_subset_sum_empty_input: (vec![], 10, None),
+        test_subset_sum_no_solution: (vec![1, 2, 3, 4, 5], 100, None),
+        test_subset_sum_single_element: (vec![7], 7, Some(vec![vec![7]])),
+        test_subset_sum_with_duplicates: (vec![1, 2, 2, 3], 5, Some(vec![vec![1, 2, 2], vec![2, 3]])),
+        test_subset_sum_with_non_unique_elements: (vec![1, 2, 3, 3, 4], 7, Some(vec![vec![1, 2, 4], vec![1, 3, 3], vec![3, 4]])),
```
vil02, line 132: Do you somehow relay on the fact that the inputs are sorted?

src/dynamic_programming/subset_sum.rs
```diff
@@ -0,0 +1,134 @@
...
+/// # Arguments
+///
+/// * `nums` - A slice of unsigned 32-bit integers representing the input numbers.
+/// * `target_sum` - The target sum to which subsets are sought.
+///
+/// # Returns
+///
+/// An option containing a vector of vectors representing subsets that sum up to the target sum,
+/// or `None` if no such subsets exist.
+pub fn subset_sum(nums: &[u32], target_sum: u32) -> Option<Vec<Vec<u32>>> {
```
vil02, line 12: It is still not clear to me what are you aiming for. If you want to handle a _general case_ (i.e. duplicates in `nums`), I would suggest to return a _set of set of indices_, for example for the input `[2, 3, 2]` and target `5`, the function should return `{{0, 1}, {0, 2}}`.

vil02 (changes requested):

src/dynamic_programming/subset_sum.rs
TruongNhanNguyen: The algorithm uses dynamic programming to generate subsets and does not assume any specific order of input elements.

src/dynamic_programming/subset_sum.rs
TruongNhanNguyen: Yeah, in some cases, when returned results for sample input `[2, 3, 2]` and target `5` are `[2, 3]` and `[3, 2]`. The code would only return `[2, 3]` (because `[3, 2]` is a duplication). Returning indices instead of numbers is a good idea. I will try it.

src/dynamic_programming/subset_sum.rs
TruongNhanNguyen: Since the table is used to store information about subsets that sum up to specific values, a more descriptive name could improve readability and maintainability of the code. I have changed the variable name from `dp` to `subset_table`.

TruongNhanNguyen: I have merged the processes of initialization and populating `subset_table` into a single function `generate_subset_table`.

src/dynamic_programming/subset_sum.rs
TruongNhanNguyen: I have updated the implementation to make `find_subset_with_sum` function (formerly `subset_sum`) return vector of vector representing indices of elements in the input `nums` vector that form subsets summing up to the target sum, or `None` if no such subsets exist.

src/dynamic_programming/subset_sum.rs
```diff
@@ -0,0 +1,128 @@
...
+            )*
+        }
+    }
+
+    test_subset_sum! {
+        test_subset_sum_basic: (vec![5, 1, 3, 4, 2], 5, Some(vec![vec![0], vec![1, 3], vec![2, 4]])),
+        test_subset_sum_empty_input: (vec![], 10, None),
+        test_subset_sum_no_solution: (vec![1, 3, 2, 5, 4], 100, None),
+        test_subset_sum_single_element: (vec![7], 7, Some(vec![vec![0]])),
+        test_subset_sum_with_duplicates: (vec![1, 3, 2, 2], 5, Some(vec![vec![0, 2, 2], vec![1, 2]])),
```
vil02, line 125:
```suggestion
        test_subset_sum_with_duplicates: (vec![1, 3, 2, 2], 5, Some(vec![vec![0, 2, 2], vec![1, 2]])),
```

There is something wrong with the implementation. You allow `vec![0, 2, 2]` but not `vec![0, 0, 0, 0, 0]`. I thin the goal was to use each entry of the input only once.

TruongNhanNguyen:
I have try some methods to reconstruct the vectors of vectors (and sets of sets) representing indices of elements in `nums` array that sum up to specific value using dynamic programming, but it not yet successful. May be implementing the subset sum using dynamic programming isn't an effective way to generate all possible vectors of indices vector (because of time and space complexity grown fast when input size increase). Other effective way to do that is using backtracking strategy.

May this pull request will be closed soon. Thanks for your carefully review @vil02.

PR #700, opened 2024-03-29
fix: suppress clippy warning

vil02:
# Pull Request Template

## Description

One of the [recent jobs](https://github.com/TheAlgorithms/Rust/actions/runs/8481497233/job/23238945730?pr=695#step:3:91) has failed due to some new clippy warning. Basically the problem is that the second coordinate of `Point` defined in:
https://github.com/TheAlgorithms/Rust/blob/1d72d9fa706af87d9e90db91140bcd6d6adc14eb/src/data_structures/heap.rs#L159
is not used.

I am aware of 3 other _fixes_ but they are all not perfect:
- use [`Point`](https://docs.rs/rusttype/latest/rusttype/struct.Point.html#) - might be an overkill in this situation,
- declare the second coordinate as `()` - _strange_,
- mark this structure as `pub` - also _strange_.

So I decided to simply suppress this warning in this line.

@siriak please have a look.

## Type of change

## Checklist:

### DIFF
--- a/src/data_structures/heap.rs
+++ b/src/data_structures/heap.rs
@@ -156,6 +156,7 @@ mod tests {
         assert_eq!(heap.pop(), Some(2));
     }
 
+    #[allow(dead_code)]
     struct Point(/* x */ i32, /* y */ i32);
 
     #[test]


siriak (approved):
Looks good, thanks!

PR #702, opened 2024-04-07
Implement Edit Distance algorithm using Dynamic Programming Strategy

TruongNhanNguyen:
## Description

Implement Edit Distance algorithm using Dynamic Programming Strategy

## Type of change

- [x] New feature (implement edit distance using dp)

## Checklist:

### DIFF
--- a/src/dynamic_programming/edit_distance.rs
+++ b/src/dynamic_programming/edit_distance.rs
@@ -1,0 +1,78 @@
+//! This module provides functions to calculate the min distance between two strings.
+//!
+//! The min distance, also known as the edit distance, is a measure of the similarity between two strings
+//! by calculating the minimum number of single-character edits (insertions, deletions, or substitutions) required
+//! to change one string into the other.
+
+/// Calculates the min distance between two strings.
+///
+/// # Arguments
+///
+/// * `string1` - A reference to the first string.
+/// * `string2` - A reference to the second string.
+///
+/// # Returns
+///
+/// The min distance between the two input strings.
+pub fn min_distance(string1: &str, string2: &str) -> usize {
+    let distance_matrix: Vec<Vec<usize>> = (0..=string1.len())
+        .map(|i| {
+            (0..=string2.len())
+                .map(|j| {
+                    if i == 0 {
+                        j
+                    } else if j == 0 {
+                        i
+                    } else {
+                        0
+                    }
+                })
+                .collect()
+        })
+        .collect();
+
+    let updated_matrix = (1..=string1.len()).fold(distance_matrix, |matrix, i| {
+        (1..=string2.len()).fold(matrix, |mut inner_matrix, j| {
+            let cost = if string1.as_bytes()[i - 1] == string2.as_bytes()[j - 1] {
+                0
+            } else {
+                1
+            };
+            inner_matrix[i][j] = (inner_matrix[i - 1][j - 1] + cost)
+                .min(inner_matrix[i][j - 1] + 1)
+                .min(inner_matrix[i - 1][j] + 1);
+            inner_matrix
+        })
+    });
+
+    updated_matrix[string1.len()][string2.len()]
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    macro_rules! min_distance_tests {
+        ($($name:ident: $test_case:expr,)*) => {
+            $(
+                #[test]
+                fn $name() {
+                    let (string1, string2, expected_distance) = $test_case;
+                    assert_eq!(
+                        min_distance(string1, string2),
+                        expected_distance
+                    );
+                }
+            )*
+        };
+    }
+
+    min_distance_tests! {
+        test_empty_strings: ("", "", 0),
+        test_same_strings: ("Hello, World!", "Hello, World!", 0),
+        test_one_empty_string: ("", "Rust", 4),
+        test_longer_first_string: ("horse", "ros", 3),
+        test_longer_second_string: ("execute", "intention", 8),
+        test_different_strings: ("elephant", "ant", 5),
+    }
+}
--- a/DIRECTORY.md
+++ b/DIRECTORY.md
@@ -77,6 +77,7 @@
   * Dynamic Programming
     * [Coin Change](https://github.com/TheAlgorithms/Rust/blob/master/src/dynamic_programming/coin_change.rs)
     * [Egg Dropping](https://github.com/TheAlgorithms/Rust/blob/master/src/dynamic_programming/egg_dropping.rs)
+    * [Edit Distance](https://github.com/TheAlgorithms/Rust/blob/master/src/dynamic_programming/edit_distance.rs)
     * [Fibonacci](https://github.com/TheAlgorithms/Rust/blob/master/src/dynamic_programming/fibonacci.rs)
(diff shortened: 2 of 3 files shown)


src/dynamic_programming/edit_distance.rs
```diff
@@ -0,0 +1,78 @@
...
+    macro_rules! min_distance_tests {
+        ($($name:ident: $test_case:expr,)*) => {
+            $(
+                #[test]
+                fn $name() {
+                    let (string1, string2, expected_distance) = $test_case;
+                    assert_eq!(
+                        min_distance(string1, string2),
+                        expected_distance
+                    );
```
vil02, line 64: I would also add few more checks like:
```suggestion
                    assert_eq!(
                        min_distance(string1, string2),
                        expected_distance
                    );
                    assert_eq!(
                        min_distance(string2, string1),
                        expected_distance
                    );
                    assert_eq!(
                        min_distance(string1, string1),
                        0
                    );
                    assert_eq!(
                        min_distance(string2, string2),
                        0
                    );
```

vil02 (commented):
Please correct me if I am wrong, but I think this is already implemented in [`levenshtein_distance.rs`](https://github.com/TheAlgorithms/Rust/blob/95bf686ceda88d9828900fde8fe7e63a1671b9aa/src/string/levenshtein_distance.rs).

Also, please notice that [Levenshtein distance](https://en.wikipedia.org/wiki/Levenshtein_distance) is an example of [edit distance](https://en.wikipedia.org/wiki/Edit_distance).

TruongNhanNguyen:
> Please correct me if I am wrong, but I think this is already implemented in [`levenshtein_distance.rs`](https://github.com/TheAlgorithms/Rust/blob/95bf686ceda88d9828900fde8fe7e63a1671b9aa/src/string/levenshtein_distance.rs).
> 
> Also, please notice that [Levenshtein distance](https://en.wikipedia.org/wiki/Levenshtein_distance) is an example of [edit distance](https://en.wikipedia.org/wiki/Edit_distance).

Yes, [Levenshtein distance](https://en.wikipedia.org/wiki/Levenshtein_distance) is an example of [edit distance](https://en.wikipedia.org/wiki/Edit_distance). Levenshtein distance was implemented in this repo but it did not use dynamic programming to solve the problem. Indeed, my implementation use dynamic programming strategy to solve it. Both implementations solve the same problem but use different approaches.

TruongNhanNguyen:
What do you think about this PR @vil02?

vil02:
The existing [`levenshtein_distance.rs`](https://github.com/TheAlgorithms/Rust/blob/95bf686ceda88d9828900fde8fe7e63a1671b9aa/src/string/levenshtein_distance.rs) used a _1D_ array to store the _partial results_. Yours is _2D_. I think both implementations have $O(nm)$ time complexity, Regarding space complexity it is $O(n)$ vs. $O(nm)$, where $n$, and $m$ are the lengths of the input strings.

I like the way how your code is written. I can see a potential to have both of implementations exist in parallel (as two different functions in one file).

TruongNhanNguyen:
I would suggest that we should keep the DP implementation of Levenshtein distance] (https://en.wikipedia.org/wiki/Levenshtein_distance) problem in the `dynamic_programming` directory for clarity of used strategy (or algorithm). It will be helpful for further updates and help us classify algorithms.

vil02 (changes requested):
@TruongNhanNguyen please use the name `levenshtein_distance` not the `edit_distance`move your file to the `string` directory and suggest a way how two implementations can exist together.

TruongNhanNguyen:
There's already a file named `levenshtein_distance` in the string directory, and that your implementation uses a different strategy. How you would like me to proceed?

TruongNhanNguyen:
What do you think about placing two files in the string directory and giving them these names: `levenshtein_distance_opt_dp` and `levenshtein_distance_dp` for the optimized version and standard version of the dynamic programming strategy?

TruongNhanNguyen:
Have a look on my suggestions, or propose a alternative solution @vil02, please.

vil02:
> Have a look on my suggestions, or propose a alternative solution @vil02, please.

Which suggestions?

TruongNhanNguyen:
> What do you think about placing two files in the `string` directory and giving them these names: `levenshtein_distance_opt_dp` and `levenshtein_distance_dp` for the optimized version and standard version of the dynamic programming strategy?
> 
> > It may be a reasonable approach! It effectively distinguishes between the two implementations based on their optimization level while still maintaining clarity about the strategy used (dynamic programming).

This is my suggestion.

vil02:
> > What do you think about placing two files in the `string` directory and giving them these names: `levenshtein_distance_opt_dp` and `levenshtein_distance_dp` for the optimized version and standard version of the dynamic programming strategy?
> > > It may be a reasonable approach! It effectively distinguishes between the two implementations based on their optimization level while still maintaining clarity about the strategy used (dynamic programming).
> 
> This is my suggestion.

Go for it!

TruongNhanNguyen:
Could you please review the updates, @vil02?

vil02:
@TruongNhanNguyen sorry for the confusion. I browsed though the repository and I think we should follow the way used in [`factorial.rs`](https://github.com/TheAlgorithms/Rust/blob/c0326771decf5bb90ad6bcf50d28f7a2fb01ef4c/src/math/factorial.rs). So there should be one file `src/string/levenshtein_distance.rs` with both implementations. Clearly the tests should be _parametrized_.

TruongNhanNguyen:
Let's have a final look, feel free to give suggestions if needed @vil02.

src/string/levenshtein_distance.rs
```diff
@@ -60,94 +130,56 @@ fn _min3<T: Ord>(a: T, b: T, c: T) -> T {
 
 #[cfg(test)]
 mod tests {
```
vil02, line 132: What do you think about organizing the tests like this:

```rust
#[cfg(test)]
mod tests {
    const LEVENSHTEIN_DISTANCE_TEST_CASES: &[(&str, &str, usize)] = &[
        ("", "", 0),
        ("Hello, World!", "Hello, World!", 0),
        ("", "Rust", 4),
        ("horse", "ros", 3),
        ("tan", "elephant", 6),
        ("execute", "intention", 8),
    ];

    macro_rules! levenshtein_distance_tests {
        ($function:ident) => {
            mod $function {
                use super::*;

                fn run_test_case(string1: &str, string2: &str, expected_distance: usize) {
                    assert_eq!(super::super::$function(string1, string2), expected_distance);
                    assert_eq!(super::super::$function(string2, string1), expected_distance);
                    assert_eq!(super::super::$function(string1, string1), 0);
                    assert_eq!(super::super::$function(string2, string2), 0);
                }

                #[test]
                fn test_levenshtein_distance() {
                    for &(string1, string2, expected_distance) in
                        LEVENSHTEIN_DISTANCE_TEST_CASES.iter()
                    {
                        run_test_case(string1, string2, expected_distance);
                    }
                }
            }
        };
    }

    levenshtein_distance_tests!(naive_levenshtein_distance);
    levenshtein_distance_tests!(optimized_levenshtein_distance);
}
```

The advantage is that the test cases are not duplicated. The disadvantage is that the test name does not contain the _current test case_, so if something fails, one will only get a message like _assertion failed in line 167_.

TruongNhanNguyen: LGTM @vil02 , thanks for the suggestions.

PR #707, opened 2024-04-28
Add `isomorphism.rs`

mengfansheng-git:
Given two strings s and t, determine whether they are isomorphic.
The two strings are isomorphic if the characters in s can be replaced by some mapping relation to get t

### DIFF
--- a/src/string/isomorphism.rs
+++ b/src/string/isomorphism.rs
@@ -72,7 +72,9 @@ mod tests {
     }
     test_is_isomorphic! {
         isomorphic: ("egg", "add", true),
+        isomorphic_long: ("abcdaabdcdbbabababacdadad", "AbCdAAbdCdbbAbAbAbACdAdAd", true),
         not_isomorphic: ("egg", "adc", false),
+        non_isomorphic_long: ("abcdaabdcdbbabababacdadad", "AACdAAbdCdbbAbAbAbACdAdAd", false),
         isomorphic_unicode: ("天苍苍", "野茫茫", true),
         isomorphic_unicode_different_byte_size: ("abb", "野茫茫", true),
         empty: ("", "", true),
--- a/src/string/isomorphism.rs
+++ b/src/string/isomorphism.rs
@@ -5,52 +5,55 @@ The two strings are isomorphic if the characters in s can be replaced by some ma
 use std::collections::HashMap;
 
 pub fn is_isomorphic(s: &str, t: &str) -> bool {
-    if s.len() != t.len() {
+    let s_chars: Vec<char> = s.chars().collect();
+    let t_chars: Vec<char> = t.chars().collect();
+    if s_chars.len() != t_chars.len() {
         return false;
     }
-    let sv: Vec<char> = s.chars().collect();
-    let tv: Vec<char> = t.chars().collect();
-    let mut sr1 = String::new();
-    let mut map1 = HashMap::new();
-    for i in 0..sv.len() {
-        if let Some(x) = map1.get(&sv[i]) {
-            sr1.push(*x);
-        } else {
-            map1.insert(sv[i], tv[i]);
-            sr1.push(tv[i]);
+    let mut s_to_t_map = HashMap::new();
+    let mut t_to_s_map = HashMap::new();
+    for (s_char, t_char) in s_chars.into_iter().zip(t_chars) {
+        if !check_mapping(&mut s_to_t_map, s_char, t_char)
+            || !check_mapping(&mut t_to_s_map, t_char, s_char)
+        {
+            return false;
         }
     }
-    let mut sr2 = String::new();
-    let mut map2 = HashMap::new();
-    for i in 0..sv.len() {
-        if let Some(x) = map2.get(&tv[i]) {
-            sr2.push(*x);
-        } else {
-            map2.insert(tv[i], sv[i]);
-            sr2.push(sv[i]);
+    true
+}
+fn check_mapping(map: &mut HashMap<char, char>, key: char, value: char) -> bool {
+    match map.get(&key) {
+        Some(&mapped_char) => mapped_char == value,
+        None => {
+            map.insert(key, value);
+            true
         }
     }
-    sr1 == t && sr2 == s
 }
 
 #[cfg(test)]
 mod tests {
     use super::is_isomorphic;
-
-    #[test]
-    fn test_is_isomorphic1() {
-        assert!(is_isomorphic("egg", "add"));
+    macro_rules! test_is_isomorphic {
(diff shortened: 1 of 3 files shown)


src/string/isomorphism.rs
```diff
@@ -0,0 +1,56 @@
...
+    for i in 0..sv.len() {
+        if let Some(x) = map2.get(&tv[i]) {
+            sr2.push(*x);
+        } else {
+            map2.insert(tv[i], sv[i]);
+            sr2.push(sv[i]);
+        }
+    }
+    sr1 == t && sr2 == s
+}
```
vil02, line 34:
```suggestion
pub fn is_isomorphic(s: &str, t: &str) -> bool {
    let s_chars: Vec<char> = s.chars().collect();
    let t_chars: Vec<char> = t.chars().collect();
    if s_chars.len() != t_chars.len() {
        return false;
    }
    let mut s_to_t_map = HashMap::new();
    let mut t_to_s_map = HashMap::new();

    for (s_char, t_char) in s_chars.into_iter().zip(t_chars) {
        if !check_mapping(&mut s_to_t_map, s_char, t_char)
            || !check_mapping(&mut t_to_s_map, t_char, s_char)
        {
            return false;
        }
    }

    true
}

fn check_mapping(map: &mut HashMap<char, char>, key: char, value: char) -> bool {
    match map.get(&key) {
        Some(&mapped_char) => mapped_char == value,
        None => {
            map.insert(key, value);
            true
        }
    }
}
```

Please note that there are few things in this suggestion:
- the _bug_ when two strings have same number of characters but different number of bytes,
- avoids need of constructing additional strings,
- reduced the code duplication.

src/string/isomorphism.rs
```diff
@@ -0,0 +1,56 @@
...
+    }
+    #[test]
+    fn test_unicode_is_isomorphic() {
+        assert!(is_isomorphic("天苍苍", "野茫茫"));
+    }
+    #[test]
+    fn test_empty_is_isomorphic() {
+        assert!(is_isomorphic("", ""));
+    }
+}
```
vil02, line 56: What do you think about reducing _code duplication_, adding some new test cases and _checks_?

```suggestion
#[cfg(test)]
mod tests {
    use super::is_isomorphic;

    macro_rules! test_is_isomorphic {
        ($($name:ident: $inputs:expr,)*) => {
        $(
            #[test]
            fn $name() {
                let (s, t, expected) = $inputs;
                assert_eq!(is_isomorphic(s, t), expected);
                assert_eq!(is_isomorphic(t, s), expected);
                assert!(is_isomorphic(s, s));
                assert!(is_isomorphic(t, t));
            }
        )*
        }
    }

    test_is_isomorphic! {
        isomorphic: ("egg", "add", true),
        not_isomorphic: ("egg", "adc", false),
        isomorphic_unicode: ("天苍苍", "野茫茫", true),
        isomorphic_unicode_different_byte_size: ("abb", "野茫茫", true),
        empty: ("", "", true),
        different_length: ("abc", "abcd", false),
    }
}
```

vil02 (changes requested):

src/string/isomorphism.rs
mengfansheng-git: The unit tests will be written later using macros

src/string/isomorphism.rs
mengfansheng-git: ok

src/string/isomorphism.rs
```diff
@@ -0,0 +1,59 @@
+/*
+Given two strings s and t, determine whether they are isomorphic.
+The two strings are isomorphic if the characters in s can be replaced by some mapping relation to get t
+*/
```
TruongNhanNguyen, line 4: I would suggest adding a file docstring here 
```suggestion
//! This module provides functionality to determine whether two strings are isomorphic.
//!
//! Two strings are considered isomorphic if the characters in one string can be replaced
//! by some mapping relation to obtain the other string.
```

src/string/isomorphism.rs
```diff
@@ -0,0 +1,59 @@
...
+use std::collections::HashMap;
+
```
TruongNhanNguyen, line 6: Add method docstring here
```suggestion

/// Determines whether two strings are isomorphic.
///
/// Two strings are considered isomorphic if the characters in one string can be replaced
/// by some mapping relation to obtain the other string.
///
/// # Arguments
///
/// * `s` - The first string.
/// * `t` - The second string.
///
/// # Returns
///
/// `true` if the strings are isomorphic, `false` otherwise.
```

src/string/isomorphism.rs
```diff
@@ -0,0 +1,59 @@
...
+    for (s_char, t_char) in s_chars.into_iter().zip(t_chars) {
+        if !check_mapping(&mut s_to_t_map, s_char, t_char)
+            || !check_mapping(&mut t_to_s_map, t_char, s_char)
+        {
+            return false;
+        }
+    }
+    true
+}
+fn check_mapping(map: &mut HashMap<char, char>, key: char, value: char) -> bool {
```
TruongNhanNguyen, line 24:
```suggestion

/// Checks the mapping between two characters and updates the map.
///
/// # Arguments
///
/// * `map` - The HashMap to store the mapping.
/// * `key` - The key character.
/// * `value` - The value character.
///
/// # Returns
///
/// `true` if the mapping is consistent, `false` otherwise.
fn check_mapping(map: &mut HashMap<char, char>, key: char, value: char) -> bool {
```

TruongNhanNguyen:
Add docstrings to clarify the purpose of functions and modules.

src/string/isomorphism.rs
TruongNhanNguyen: Yeah, sorry for the confusion, just remove duplicated lines.

src/string/isomorphism.rs
```diff
@@ -0,0 +1,81 @@
...
+                assert_eq!(is_isomorphic(t, s), expected);
+                assert!(is_isomorphic(s, s));
+                assert!(is_isomorphic(t, t));
+            }
+        )*
+        }
+    }
+    test_is_isomorphic! {
+        isomorphic: ("egg", "add", true),
+        not_isomorphic: ("egg", "adc", false),
```
vil02, line 75:
```suggestion
        isomorphic: ("egg", "add", true),
        isomorphic_long: ("abcdaabdcdbbabababacdadad", "AbCdAAbdCdbbAbAbAbACdAdAd", true),
        not_isomorphic: ("egg", "adc", false),
        non_isomorphic_long: ("abcdaabdcdbbabababacdadad", "AACdAAbdCdbbAbAbAbACdAdAd", false),
```

vil02 (approved):

PR #710, opened 2024-05-02
Implements Ramer-Douglas-Peucker simplification algorithm

Sellig6792:
# Pull Request Template

## Description

I implemented the Ramer-Douglas-Peucker algorithm. I use it in a project it works very well. This algorithm allow you to reduce the number of point of a polygon and to keep a maximum of precision.

The test doesn't make a real sesnse, since the results changes with the epsilon value.

- https://en.wikipedia.org/wiki/Ramer%E2%80%93Douglas%E2%80%93Peucker_algorithm 
- https://fr.wikipedia.org/wiki/Algorithme_de_Douglas-Peucker (FR)

## Type of change

## Checklist:

- [x] I ran bellow commands using the latest version of **rust nightly**. => My code doesn't make issues

### DIFF
--- a/src/geometry/ramer_douglas_peucker.rs
+++ b/src/geometry/ramer_douglas_peucker.rs
@@ -28,7 +28,7 @@ pub fn ramer_douglas_peucker(points: &[Point], epsilon: f64) -> Vec<Point> {
 
 fn perpendicular_distance(p: &Point, a: &Point, b: &Point) -> f64 {
     let num = (b.y - a.y) * p.x - (b.x - a.x) * p.y + b.x * a.y - b.y * a.x;
-    let den = a.euclidean_distance(&b);
+    let den = a.euclidean_distance(b);
     num.abs() / den
 }
 
--- a/src/geometry/ramer_douglas_peucker.rs
+++ b/src/geometry/ramer_douglas_peucker.rs
@@ -1,0 +1,73 @@
+use crate::geometry::Point;
+
+pub fn ramer_douglas_peucker(polygon: &[Point], epsilon: f64) -> Vec<Point> {
+    let mut dmax = 0.0;
+    let mut index = 0;
+    let end = polygon.len() - 1;
+
+    for i in 1..end {
+        let d =
+            perpendicular_distance(polygon[i].clone(), polygon[0].clone(), polygon[end].clone());
+        if d > dmax {
+            index = i;
+            dmax = d;
+        }
+    }
+
+    if dmax > epsilon {
+        let mut results = ramer_douglas_peucker(&Vec::from(&polygon[..=index]), epsilon);
+        results.pop();
+        results.extend(ramer_douglas_peucker(
+            &Vec::from(&polygon[index..]),
+            epsilon,
+        ));
+        results
+    } else {
+        vec![polygon[0].clone(), polygon[end].clone()]
+    }
+}
+
+fn perpendicular_distance(p: Point, a: Point, b: Point) -> f64 {
+    let num = (b.y - a.y) * p.x - (b.x - a.x) * p.y + b.x * a.y - b.y * a.x;
+    let den = a.euclidean_distance(&b);
+    num.abs() / den
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_perpendicular_distance() {
+        let a = Point::new(0.0, 0.0);
+        let b = Point::new(0.0, 3.0);
+        let p = Point::new(4.0, 0.0);
+        assert_eq!(perpendicular_distance(p, a, b), 4.0);
+    }
+
+    #[test]
+    fn test_ramer_douglas_peucker() {
+        // This test doesn't make sense, once you change the epsilon value, the result will change. It is hard to predict the result.
+        let a = Point::new(0.0, 0.0);
+        let b = Point::new(1.0, 0.0);
+        let c = Point::new(2.0, 0.0);
+        let d = Point::new(2.0, 1.0);
+        let e = Point::new(2.0, 2.0);
+        let f = Point::new(1.0, 2.0);
+        let g = Point::new(0.0, 2.0);
+        let h = Point::new(0.0, 1.0);
+        let polygon = vec![
+            a.clone(),
+            b,
+            c.clone(),
+            d,
+            e.clone(),
+            f,
+            g.clone(),
+            h.clone(),
+        ];
+        let epsilon = 0.7;
+        let result = ramer_douglas_peucker(&polygon, epsilon);
+        assert_eq!(result, vec![a, c, e, g, h]);
+    }
+}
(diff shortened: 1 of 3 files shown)


src/geometry/ramer_douglas_peucker.rs
```diff
@@ -0,0 +1,73 @@
+use crate::geometry::Point;
+
+pub fn ramer_douglas_peucker(polygon: &[Point], epsilon: f64) -> Vec<Point> {
```
vil02, line 3: Do you really mean `polygon`? I thought this should work for any _polygonal chain_.

src/geometry/ramer_douglas_peucker.rs
```diff
@@ -0,0 +1,73 @@
...
+    let mut dmax = 0.0;
+    let mut index = 0;
+    let end = polygon.len() - 1;
+
+    for i in 1..end {
+        let d =
+            perpendicular_distance(polygon[i].clone(), polygon[0].clone(), polygon[end].clone());
```
vil02, line 10:
```suggestion
        let d = perpendicular_distance(&polygon[i], &polygon[0], &polygon[end]);
```

src/geometry/ramer_douglas_peucker.rs
```diff
@@ -0,0 +1,73 @@
...
+        if d > dmax {
+            index = i;
+            dmax = d;
+        }
+    }
+
+    if dmax > epsilon {
+        let mut results = ramer_douglas_peucker(&Vec::from(&polygon[..=index]), epsilon);
```
vil02, line 18:
```suggestion
        let mut results = ramer_douglas_peucker(&polygon[..=index], epsilon);
```

src/geometry/ramer_douglas_peucker.rs
```diff
@@ -0,0 +1,73 @@
...
+            &Vec::from(&polygon[index..]),
+            epsilon,
+        ));
+        results
+    } else {
+        vec![polygon[0].clone(), polygon[end].clone()]
+    }
+}
+
+fn perpendicular_distance(p: Point, a: Point, b: Point) -> f64 {
```
vil02, line 30:
```suggestion
fn perpendicular_distance(p: &Point, a: &Point, b: &Point) -> f64 {
```

src/geometry/ramer_douglas_peucker.rs
```diff
@@ -0,0 +1,73 @@
...
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_perpendicular_distance() {
+        let a = Point::new(0.0, 0.0);
+        let b = Point::new(0.0, 3.0);
+        let p = Point::new(4.0, 0.0);
+        assert_eq!(perpendicular_distance(p, a, b), 4.0);
```
vil02, line 45:
```suggestion
        assert_eq!(perpendicular_distance(&p, &a, &b), 4.0);
```

src/geometry/ramer_douglas_peucker.rs
```diff
@@ -0,0 +1,73 @@
...
+    }
+
+    #[test]
+    fn test_ramer_douglas_peucker() {
+        // This test doesn't make sense, once you change the epsilon value, the result will change. It is hard to predict the result.
```
vil02, line 50: ?

src/geometry/ramer_douglas_peucker.rs
```diff
@@ -0,0 +1,73 @@
+use crate::geometry::Point;
+
+pub fn ramer_douglas_peucker(polygon: &[Point], epsilon: f64) -> Vec<Point> {
+    let mut dmax = 0.0;
+    let mut index = 0;
+    let end = polygon.len() - 1;
```
vil02, line 6: What should happen if the input line it _too short_?

vil02 (changes requested):

src/geometry/ramer_douglas_peucker.rs
Sellig6792: Yes it does, you are right, should i name it `points`?

src/geometry/ramer_douglas_peucker.rs
Sellig6792: I just mean that the result changes with the epsilon so you can't just predict the output easily

src/geometry/ramer_douglas_peucker.rs
```diff
@@ -0,0 +1,86 @@
+use crate::geometry::Point;
+
+pub fn ramer_douglas_peucker(points: &[Point], epsilon: f64) -> Vec<Point> {
+    if points.len() < 3 {
+        return points.to_vec();
+    }
```
vil02, line 6: Please add tests, where you call `ramer_douglas_peucker` with empty, single point and two point inputs.

src/geometry/ramer_douglas_peucker.rs
```diff
@@ -0,0 +1,86 @@
...
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_perpendicular_distance() {
+        let a = Point::new(0.0, 0.0);
+        let b = Point::new(0.0, 3.0);
+        let p = Point::new(4.0, 0.0);
+        assert_eq!(perpendicular_distance(&p, &a, &b), 4.0);
+    }
```
vil02, line 45:
```suggestion
    macro_rules! test_perpendicular_distance {
        ($($name:ident: $test_case:expr,)*) => {
            $(
                #[test]
                fn $name() {
                    let (p, a, b, expected) = $test_case;
                    assert_eq!(perpendicular_distance(&p, &a, &b), expected);
                    assert_eq!(perpendicular_distance(&p, &b, &a), expected);
                }
            )*
        };
    }

    test_perpendicular_distance! {
        basic: (Point::new(4.0, 0.0), Point::new(0.0, 0.0), Point::new(0.0, 3.0), 4.0),
        basic_shifted_1: (Point::new(4.0, 1.0), Point::new(0.0, 1.0), Point::new(0.0, 4.0), 4.0),
        basic_shifted_2: (Point::new(2.0, 1.0), Point::new(-2.0, 1.0), Point::new(-2.0, 4.0), 4.0),
    }
```

src/geometry/ramer_douglas_peucker.rs
```diff
@@ -0,0 +1,86 @@
...
+            g.clone(),
+            h.clone(),
+        ];
+        let epsilon = 0.7;
+        let result = ramer_douglas_peucker(&polygon, epsilon);
+        assert_eq!(result, vec![a, c, e, g, h]);
+    }
+
+    #[test]
+    fn test_polygon_chain() {
```
vil02, line 73: Please change the name of this test.

vil02 (changes requested):

src/geometry/ramer_douglas_peucker.rs
```diff
@@ -0,0 +1,115 @@
...
+        if d > dmax {
+            index = i;
+            dmax = d;
+        }
+    }
+
+    if dmax > epsilon {
+        let mut results = ramer_douglas_peucker(&points[..=index], epsilon);
+        results.pop();
+        results.extend(ramer_douglas_peucker(&Vec::from(&points[index..]), epsilon));
```
vil02, line 22:
```suggestion
        results.extend(ramer_douglas_peucker(&points[index..], epsilon));
```

vil02 (approved):

PR #711, opened 2024-05-06
add check_permutation.rs

mengfansheng-git:
Given two strings s1 and s2.
Determine whether the characters of one of the strings can be rearranged to become another string.

### DIFF
--- a/src/string/check_permutation.rs
+++ b/src/string/check_permutation.rs
@@ -1,0 +1,40 @@
+//! Given two strings s1 and s2 consisting of lowercase letters.
+//!
+//! Determine whether the characters of one of the strings can be
+//! rearranged to become another string.
+
+pub fn check_permutation(s1: &str, s2: &str) -> bool {
+    if s1.len() != s2.len() {
+        return false;
+    }
+    s1.chars()
+        .all(|c| s1.matches(c).count() == s2.matches(c).count())
+}
+
+#[cfg(test)]
+mod tests {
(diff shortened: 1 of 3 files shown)


src/string/check_permutation.rs
```diff
@@ -0,0 +1,40 @@
+//! Given two strings s1 and s2 consisting of lowercase letters.
```
vil02, line 1: What if they are not lowercase letters?

src/string/check_permutation.rs
```diff
@@ -0,0 +1,40 @@
...
+//!
+//! Determine whether the characters of one of the strings can be
+//! rearranged to become another string.
+
+pub fn check_permutation(s1: &str, s2: &str) -> bool {
+    if s1.len() != s2.len() {
+        return false;
+    }
+    s1.chars()
+        .all(|c| s1.matches(c).count() == s2.matches(c).count())
```
vil02, line 11: What is the time complexity of this? $O(n^2)$? I think one can do it in a _linear_ time.

vil02 (changes requested):

src/string/check_permutation.rs
mengfansheng-git: have no effect

src/string/check_permutation.rs
mengfansheng-git: With the help of a hash table to record the number of occurrences of a character and then compare the time complexity should be linear time.

PR #713, opened 2024-05-12
Implement Parentheses Generator using Backtracking Strategy

TruongNhanNguyen:
## Type of change

## Checklist:

### DIFF
--- a/src/backtracking/parentheses_generator.rs
+++ b/src/backtracking/parentheses_generator.rs
@@ -1,11 +1,11 @@
-/// Generates all combinations of well-formed parentheses given a non-positive integer `n`.
+/// Generates all combinations of well-formed parentheses given a non-negative integer `n`.
 ///
 /// This function uses backtracking to generate all possible combinations of well-formed
 /// parentheses. The resulting combinations are returned as a vector of strings.
 ///
 /// # Arguments
 ///
-/// * `n` - A non-positive integer representing the number of pairs of parentheses.
+/// * `n` - A non-negative integer representing the number of pairs of parentheses.
 pub fn generate_parentheses(n: usize) -> Vec<String> {
     if n == 0 {
         return Vec::new();
--- a/src/backtracking/parentheses_generator.rs
+++ b/src/backtracking/parentheses_generator.rs
@@ -25,21 +25,25 @@ pub fn generate_parentheses(n: usize) -> Vec<String> {
 /// * `close_count` - The count of close parentheses in the current string.
 /// * `n` - The total number of pairs of parentheses to be generated.
 /// * `result` - A mutable reference to the vector storing the generated combinations.
-fn generate(current: &str, open_count: usize, close_count: usize, n: usize, result: &mut Vec<String>) {
+fn generate(
+    current: &str,
+    open_count: usize,
+    close_count: usize,
+    n: usize,
+    result: &mut Vec<String>,
+) {
     if current.len() == (n * 2) {
         result.push(current.to_string());
         return;
     }
 
     if open_count < n {
-        let mut new_str = current.to_string();
-        new_str.push('(');
+        let new_str = current.to_string() + "(";
         generate(&new_str, open_count + 1, close_count, n, result);
     }
 
     if close_count < open_count {
-        let mut new_str = current.to_string();
-        new_str.push(')');
+        let new_str = current.to_string() + ")";
         generate(&new_str, open_count, close_count + 1, n, result);
     }
 }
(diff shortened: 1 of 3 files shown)


src/backtracking/parentheses_generator.rs
```diff
@@ -0,0 +1,69 @@
...
+/// * `result` - A mutable reference to the vector storing the generated combinations.
+fn generate(current: &str, open: u32, close: u32, n: u32, result: &mut Vec<String>) {
+    if current.len() == (n * 2) as usize {
+        result.push(current.to_string());
+        return;
+    }
+
+    if open < n {
+        let mut new_str = current.to_string();
+        new_str.push('(');
```
vil02, line 36:
```suggestion
        let new_str = current.to_string() + "(";
        new_str.push('(');
```

src/backtracking/parentheses_generator.rs
```diff
@@ -0,0 +1,69 @@
...
+        generate(&new_str, open + 1, close, n, result);
+    }
+
+    if close < open {
+        let mut new_str = current.to_string();
+        new_str.push(')');
```
vil02, line 42:
```suggestion
        let new_str = current.to_string() + ")";
```

src/backtracking/parentheses_generator.rs
```diff
@@ -0,0 +1,69 @@
...
+/// if it's valid to do so.
+///
+/// # Arguments
+///
+/// * `current` - The current string of parentheses being built.
+/// * `open` - The count of open parentheses in the current string.
+/// * `close` - The count of close parentheses in the current string.
+/// * `n` - The total number of pairs of parentheses to be generated.
+/// * `result` - A mutable reference to the vector storing the generated combinations.
+fn generate(current: &str, open: u32, close: u32, n: u32, result: &mut Vec<String>) {
```
vil02, line 28: What do you think about renaming `open` to `open_count` etc?

src/backtracking/parentheses_generator.rs
```diff
@@ -0,0 +1,69 @@
+/// Generates all combinations of well-formed parentheses given a positive integer `n`.
+///
+/// This function uses backtracking to generate all possible combinations of well-formed
+/// parentheses. The resulting combinations are returned as a vector of strings.
+///
+/// # Arguments
+///
+/// * `n` - A positive integer representing the number of pairs of parentheses.
+pub fn generate_parentheses(n: u32) -> Vec<String> {
```
vil02, line 9: I would use `usize` as a type of `n`, because later on, you cast it to `usize` anyway.

vil02 (changes requested):

src/backtracking/parentheses_generator.rs
TruongNhanNguyen: It's a good idea. I will change `open` and `close` to `open_count` and `close_count` respectively.

src/backtracking/parentheses_generator.rs
TruongNhanNguyen: It should look like this:
Old:
```rust
let mut new_str = current.to_string();
new_str.push('(');
```
New:
```rust
let new_str = current.to_string() + "(";
```

src/backtracking/parentheses_generator.rs
```diff
@@ -0,0 +1,73 @@
...
+                fn $name() {
+                    let (n, expected_result) = $test_case;
+                    assert_eq!(generate_parentheses(n), expected_result);
+                }
+            )*
+        };
+    }
+
+    generate_parentheses_tests! {
+        test_generate_parentheses_1: (1, vec!["()"]),
```
vil02, line 68: I missed that earlier: I think one should test it with `0` as an input.

I am not sure what should be the input. This has some relationship to Catalan numbers (and passes with current implementation):
```suggestion
        test_generate_parentheses_0: (0, vec![""]),
        test_generate_parentheses_1: (1, vec!["()"]),
```
This is more _intuitive_:
```suggestion
        test_generate_parentheses_0: (0, Vec::<String>::new()),
        test_generate_parentheses_1: (1, vec!["()"]),
```

vil02 (changes requested):

src/backtracking/parentheses_generator.rs
TruongNhanNguyen: Thanks

src/backtracking/parentheses_generator.rs
```diff
@@ -0,0 +1,77 @@
...
+///
+/// # Arguments
+///
+/// * `n` - A non-negative integer representing the number of pairs of parentheses.
+pub fn generate_parentheses(n: usize) -> Vec<String> {
+    if n == 0 {
+        return Vec::new();
+    }
+    let mut result = Vec::new();
+    generate("", 0, 0, n, &mut result);
```
vil02, line 14:
```suggestion
    let mut result = Vec::new();
    if n > 0 {
        generate("", 0, 0, n, &mut result);
    }
```

vil02 (approved):

PR #720, opened 2024-05-17
Refactor Caesar Algorithm Implementation

StPfeffer:
## Description

Refactored the Caesar algorithm to handle larger inputs and rotations while preserving the existing code structure. Included additional tests for improved coverage.

## Type of change

## Checklist:


src/ciphers/caesar.rs
```diff
@@ -1,20 +1,25 @@
...
+///
+/// # Arguments
+///
+/// * `text` - The text to be encrypted.
+/// * `rotation` - The number of rotations (shift) to be applied.
+///
+/// # Returns
+///
+/// Returns the encrypted string.
+pub fn caesar(text: &str, rotation: usize) -> String {
```
vil02, line 16: I would check if `0 <= rotation < 26`, because using any other rotation does not make too much sense.

src/ciphers/caesar.rs
```diff
@@ -1,20 +1,25 @@
...
+    text.chars()
         .map(|c| {
             if c.is_ascii_alphabetic() {
                 let first = if c.is_ascii_lowercase() { b'a' } else { b'A' };
-                // modulo the distance to keep character range
-                (first + (c as u8 + shift - first) % 26) as char
+
+                (((c as u8 - first) + (rotation % 26) as u8) % 26 + first) as char
             } else {
                 c
```
vil02, line 24: Since this is the essence of the entire cipher, I would move it into a separate function like `rotate_char` or `shift_char`. What do you think?

src/ciphers/caesar.rs
```diff
@@ -32,12 +37,92 @@ mod tests {
     }
```
vil02, line 38: What do you think about expressing all of the tests as follows:

```rust
#[cfg(test)]
mod tests {
    use super::*;
    macro_rules! test_caesar {
            $(
                #[test]
                fn $name() {
                    let (text, rotation, expected) = $test_case;
                    assert_eq!(caesar(&text, rotation), expected);
                    assert_eq!(caesar(&expected, 26-rotation), text);
                }
            )*
        };
    }

    test_caesar! {
        empty_text: ("", 13, ""),
        rot_13: ("rust", 13, "ehfg"),
        unicode: ("attack at dawn 攻", 5, "fyyfhp fy ifbs 攻"),
        rotation_within_alphabet_range: ("Hello, World!", 3, "Khoor, Zruog!"),
        no_rotation: ("Hello, World!", 0, "Hello, World!"),
        rotation_at_alphabet_end: ("Hello, World!", 25, "Gdkkn, Vnqkc!"),
        longer: ("The quick brown fox jumps over the lazy dog.", 5, "Ymj vznhp gwtbs ktc ozrux tajw ymj qfed itl."),
        non_alphabetic_characters: ("12345!@#$%", 3, "12345!@#$%"),
        uppercase_letters: ("ABCDEFGHIJKLMNOPQRSTUVWXYZ", 1, "BCDEFGHIJKLMNOPQRSTUVWXYZA"),
        mixed_case: ("HeLlO WoRlD", 7, "OlSsV DvYsK"),
        with_whitespace: ("Hello, World!", 13, "Uryyb, Jbeyq!"),
        with_special_characters: ("Hello!@#$%^&*()_+World", 4, "Lipps!@#$%^&*()_+Asvph"),
        with_numbers: ("Abcd1234XYZ", 10, "Klmn1234HIJ"),
    }
}
```

Please note that I kipped the _long inputs_ and _big rotation_.

vil02 (changes requested):

src/ciphers/caesar.rs
StPfeffer: That's a really good way to do the tests, thanks for pointing it out!

src/ciphers/caesar.rs
StPfeffer: That makes sense. What do you think about returning an error if the rotation is not within the range (0 - 25), changing the return type to `Result<String, &'static str>`?

src/ciphers/caesar.rs
StPfeffer: Great idea, this will make the function much more readable. Consider it done.

src/ciphers/caesar.rs
vil02: Go for it. Maybe even without explicitly defining `ERROR_MESSAGE`.

src/ciphers/caesar.rs
```diff
@@ -1,43 +1,117 @@
...
+                        _ => {
+                            // If rotation is invalid, test that the function returns the specific error message
+                            match caesar(&text, rotation) {
+                                Ok(_) => panic!("Expected an error but got an Ok result"),
+                                Err(e) => assert_eq!(e, ERROR_MESSAGE),
+                            }
+                        }
+                    }
+                }
+            )*
```
vil02, line 94: This is way to complex. Please split it into two tests: one checking the _happy path_ (something like [suggested here](https://github.com/TheAlgorithms/Rust/pull/720#discussion_r1606048554)) and the second, with inputs, for which `caesar` returns an error.

StPfeffer: Created the following macro_rules: `test_caesar_happy_path` and `test_caesar_error_cases`.

I also updated the tests to match with the new rules.

src/ciphers/caesar.rs
```diff
@@ -1,43 +1,126 @@
...
+
+                    // If rotation is invalid, test that the function returns the specific error message
+                    match caesar(&text, rotation) {
+                        Ok(_) => panic!("Expected an error but got an Ok result"),
+                        Err(e) => assert_eq!(e, ERROR_MESSAGE),
+                    }
+                }
+            )*
+        };
     }
```
vil02, line 102:
```suggestion
    macro_rules! test_caesar_happy_path {
        ($($name:ident: $test_case:expr,)*) => {
            $(
                #[test]
                fn $name() {
                    let (text, rotation, expected) = $test_case;
                    assert_eq!(caesar(&text, rotation).unwrap(), expected);

                    let backward_rotation = if rotation == 0 { 0 } else { 26 - rotation };
                    assert_eq!(caesar(&expected, backward_rotation).unwrap(), text);
                }
            )*
        };
    }

    macro_rules! test_caesar_error_cases {
        ($($name:ident: $test_case:expr,)*) => {
            $(
                #[test]
                fn $name() {
                    let (text, rotation) = $test_case;
                    assert_eq!(caesar(&text, rotation), Err(ERROR_MESSAGE));
                }
            )*
        };
    }
```

src/ciphers/caesar.rs
```diff
@@ -1,43 +1,126 @@
...
+    test_caesar_error_cases! {
+        negative_rotation: ("Hello, World!", -5),
+        large_rotation: ("Large rotation", 139),
+        large_rotation_with_big_input: ("Lorem ipsum dolor sit amet, consectetur adipiscing elit. Curabitur ac ultrices ante, at gravida ante. Quisque luctus, ligula nec dictum facilisis, elit leo luctus arcu, ut auctor sapien turpis ut mauris. Orci varius natoque penatibus et magnis dis parturient montes, nascetur ridiculus mus. Nulla vel orci sit amet sem efficitur sagittis a quis augue. Donec semper quam tincidunt hendrerit cursus. Duis placerat gravida diam, in interdum purus dapibus in.", 139),
+        very_large_rotation_with_big_input: ("Lorem ipsum dolor sit amet, consectetur adipiscing elit. Curabitur ac ultrices ante, at gravida ante. Quisque luctus, ligula nec dictum facilisis, elit leo luctus arcu, ut auctor sapien turpis ut mauris. Orci varius natoque penatibus et magnis dis parturient montes, nascetur ridiculus mus. Nulla vel orci sit amet sem efficitur sagittis a quis augue. Donec semper quam tincidunt hendrerit cursus. Duis placerat gravida diam, in interdum purus dapibus in.", 345876),
```
vil02, line 124:
```suggestion
        empty_input_negative_rotation: ("", -1),
        empty_input_large_rotation: ("", 27), 
```

vil02 (changes requested):
What do you think about defining _the_ `26` as some constant. It could be computed as a difference between `a` and `z` (plus 1, I think).

StPfeffer:
> What do you think about defining _the_ `26` as some constant. It could be computed as a difference between `a` and `z` (plus 1, I think).

That's a good idea. I defined it as:

```rust
const ALPHABET_LENGTH: u8 = b'z' - b'a' + 1;
```

src/ciphers/caesar.rs
```diff
@@ -1,43 +1,115 @@
...
-//! # Algorithm
-//!
-//! Rotate each ascii character by shift. The most basic example is ROT 13, which rotates 'a' to
-//! 'n'. This implementation does not rotate unicode characters.
-
-/// Caesar cipher to rotate cipher text by shift and return an owned String.
-pub fn caesar(cipher: &str, shift: u8) -> String {
-    cipher
+const ERROR_MESSAGE: &str = "Rotation must be in the range [0, 25]";
+const ALPHABET_LENGTH: u8 = b'z' - b'a' + 1;
```
vil02, line 2: Do you know some _reasonable_ way to check if `ALPHABET_LENGTH` is indeed `26`?

src/ciphers/caesar.rs
```diff
@@ -1,43 +1,115 @@
...
+        empty_input_negative_rotation: ("", -1),
+        empty_input_large_rotation: ("", 27),
+        large_rotation: ("Large rotation", 139),
+        large_rotation_with_big_input: ("Lorem ipsum dolor sit amet, consectetur adipiscing elit. Curabitur ac ultrices ante, at gravida ante. Quisque luctus, ligula nec dictum facilisis, elit leo luctus arcu, ut auctor sapien turpis ut mauris. Orci varius natoque penatibus et magnis dis parturient montes, nascetur ridiculus mus. Nulla vel orci sit amet sem efficitur sagittis a quis augue. Donec semper quam tincidunt hendrerit cursus. Duis placerat gravida diam, in interdum purus dapibus in.", 139),
+        very_large_rotation_with_big_input: ("Lorem ipsum dolor sit amet, consectetur adipiscing elit. Curabitur ac ultrices ante, at gravida ante. Quisque luctus, ligula nec dictum facilisis, elit leo luctus arcu, ut auctor sapien turpis ut mauris. Orci varius natoque penatibus et magnis dis parturient montes, nascetur ridiculus mus. Nulla vel orci sit amet sem efficitur sagittis a quis augue. Donec semper quam tincidunt hendrerit cursus. Duis placerat gravida diam, in interdum purus dapibus in.", 345876),
```
vil02, line 113: Why do you insist on these two?

src/ciphers/caesar.rs
```diff
@@ -1,43 +1,115 @@
...
+///
+/// * `c` - The ASCII alphabetic character to be shifted.
+/// * `rotation` - The number of positions to shift the character. Should be within the range (0, 25).
+///
+/// # Returns
+///
+/// Returns the shifted ASCII alphabetic character.
+fn shift_char(c: char, rotation: isize) -> char {
+    let first = if c.is_ascii_lowercase() { b'a' } else { b'A' };
+    let rotation = rotation as u8; // Safe cast as rotation is within (0, 25)
```
vil02, line 55:
```suggestion
    let rotation = rotation as u8; // Safe cast as rotation is within [0, 25]
```

src/ciphers/caesar.rs
```diff
@@ -1,43 +1,115 @@
...
+
+    Ok(result)
+}
+
+/// Shifts a single ASCII alphabetic character by a specified number of positions in the alphabet.
+///
+/// # Arguments
+///
+/// * `c` - The ASCII alphabetic character to be shifted.
+/// * `rotation` - The number of positions to shift the character. Should be within the range (0, 25).
```
vil02, line 48:
```suggestion
/// * `rotation` - The number of positions to shift the character. Should be within the range [0, 25].
```

src/ciphers/caesar.rs
```diff
@@ -1,43 +1,115 @@
...
+/// * `rotation` - The number of rotations (shift) to be applied. It should be within the range (0, 25).
+///
+/// # Returns
+///
+/// Returns a `Result` containing the encrypted string if successful, or an error message if the rotation
+/// is out of the valid range.
+///
+/// # Errors
+///
+/// Returns an error if the rotation value is out of the valid range (0, 25)
```
vil02, line 23:
```suggestion
/// Returns an error if the rotation value is out of the valid range [0, 25]
```

vil02 (changes requested):

src/ciphers/caesar.rs
StPfeffer: Because the previous implementation failed to pass these two tests

src/ciphers/caesar.rs
StPfeffer: Only direct comparison in the code

vil02: This has to be placed inside a test. What do you think about simply adding a suitable test with
```rust
assert_eq!(ALPHABET_LENGTH, 26);
```

src/ciphers/caesar.rs
vil02: Was is failing due to the length of the input or size of the rotation?

StPfeffer: Rotation size

StPfeffer: Ok, removed

vil02 (approved):

PR #723, opened 2024-05-21
K nearest neighbors Algorithm

haswanth10:
This commit adds K-nearest neighbors Algorithm.

 Important points regarding Commit:
   1. Euclidean distance is used to compute the distance between 2 points.
   2. Only classification has been implemented by using majority vote amongst k nearest neighbors for an input point
   3. An Enum has been created for distance computation so that it can be extended to other distance computation techniques/formulae like Manhattan Distance, etc.
   4. Classification Label is of type 'String'.

# Pull Request Template

## Description

## Type of change

- [ Y] New feature (non-breaking change which adds functionality)

## Checklist:

- [ Y] I ran bellow commands using the latest version of **rust nightly**.
- [ Y] I ran `cargo clippy --all -- -D warnings` just before my last commit and fixed any issue that was found.
- [ Y] I ran `cargo fmt` just before my last commit.
- [Y ] I ran `cargo test` just before my last commit and all tests passed.
- [ Y] I added my algorithm to the corresponding `mod.rs` file within its own folder, and in any parent folder(s).
- [ Y] I added my algorithm to `DIRECTORY.md` with the correct link.
- [ Y] I checked `COUNTRIBUTING.md` and my code follows its guidelines.

### DIFF
--- a/src/machine_learning/optimization/k_nearest_neighbors.rs
+++ b/src/machine_learning/optimization/k_nearest_neighbors.rs
@@ -80,7 +80,6 @@ fn fetch_k_nearest_neighbors<'a>(
     distance_computation: DistanceMeasurementFormula,
 ) -> Vec<&'a Point> {
     //calculate the distance to all neighbors from input_point.
-    //let mut distance_vec: Vec<f64> = calculate_distance_to_neighbors(&neighbors, &input_point);
     let distance_map: BTreeMap<i64, &Point> =
         calculate_distance_to_neighbors(&neighbors, &input_point, distance_computation);
 
@@ -95,14 +94,12 @@ fn calculate_distance_to_neighbors<'a>(
     input_point: &'a Point,
     distance_computation: DistanceMeasurementFormula,
 ) -> BTreeMap<i64, &'a Point> {
-    // when distance is calculated , we can have multiple nodes having same distance
-    // let mut distance_vec: Vec<f64> = Vec::new();
     let mut distance_map: BTreeMap<i64, &Point> = BTreeMap::new();
 
     for i in 0..neighbors.len() {
         let neighbor = &neighbors[i];
         let distance: f64 = distance_computation.distance(&input_point, &neighbor);
-        //multiplying by 100 since floating point numbers cant be keys anymore. Any precission more than 2 digits is ignored
+        //multiplying by 100 since floating point numbers cant be keys. Any precission more than 2 digits is ignored
         distance_map.insert((distance * 100.0) as i64, &neighbor);
     }
     distance_map
--- a/src/machine_learning/optimization/k_nearest_neighbors.rs
+++ b/src/machine_learning/optimization/k_nearest_neighbors.rs
@@ -63,7 +63,7 @@ fn find_majority_label(k_nearest_neighbors: Vec<&Point>) -> String {
         }
         if counter > max_counter {
             max_counter = counter;
-            label = neighbor.label.to_owned();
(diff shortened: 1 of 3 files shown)


src/machine_learning/optimization/k_nearest_neighbors.rs
```diff
@@ -0,0 +1,294 @@
...
+    }
+
+    fn eucledian_distance(source_point: &Point, destination_point: &Point) -> f64 {
+        let distance = square_root(
+            (destination_point.x - source_point.x) * (destination_point.x - source_point.x)
+                + (destination_point.y - source_point.y) * (destination_point.y - source_point.y),
+        );
+        abs(distance)
+    }
+}
```
vil02, line 41: Why not to use [existing implementation](https://github.com/TheAlgorithms/Rust/blob/c2009cd31ea0d5171ce1c12b4c3742a2676cd07e/src/geometry/point.rs#L25)?

src/machine_learning/optimization/k_nearest_neighbors.rs
```diff
@@ -0,0 +1,294 @@
...
+    cmp,
+    collections::{BTreeMap, HashMap},
+};
+
+use crate::math::{abs, square_root};
+#[derive(Debug, Clone)]
+pub struct Point {
+    x: f64,        // x-axis
+    y: f64,        // y-axis
+    label: String, // A label denotes the classification of the data. For instance, 'Football' might be labeled as 'Sport'
```
vil02, line 14: These comments do not provide any additional value and can be removed, since.

Why do you restrict yourself to 2D data?

src/machine_learning/optimization/k_nearest_neighbors.rs
```diff
@@ -0,0 +1,294 @@
...
+        );
+        abs(distance)
+    }
+}
+
+pub fn classify_using_knn(
+    neighbors: Vec<&Point>, // The training data which essentially consists of a set of points on the X-Y axis represented as vector of Points,
+    input_point: Point,     // The input point requiring classification
+    k: usize, // The value of 'K'. For example, if K equals 4, classification is determined by the majority vote among the 4 nearest neighbors
+    distance_computation: DistanceMeasurementFormula, // An enum employed to specify the technique/formula for calculating the distance between two points
```
vil02, line 47: Why not do make it a _function pointer_, so the caller could define their own _distance_ without modifying your code?

vil02 (changes requested):

src/machine_learning/optimization/k_nearest_neighbors.rs
haswanth10: We can. However , i have changed my code to use n-dimensional data. The existing implementation is only for 2D data.

src/machine_learning/optimization/k_nearest_neighbors.rs
haswanth10: Yeah , I agree. I started with 2D data. I updated the code now to support n-dimensions data.

src/machine_learning/optimization/k_nearest_neighbors.rs
haswanth10: Yeah, I was also thinking about using traits to achieve the same. But i assume, traits do add a vtable's lookup overhead which can be avoided with function pointers. I updated the code with the same

haswanth10:
Hi , i made the required changes. Let me know if there is anything else that's needed :)

PR #725, opened 2024-05-24
add maximum.rs

mengfansheng-git:
Calculate the maximum set of multiplications of three numbers in an array

### DIFF
--- a/src/big_integer/maximum.rs
+++ b/src/big_integer/maximum.rs
@@ -7,9 +7,9 @@
 //         .max(nums[0] * nums[1] * nums[nums.len() - 1])
 // }
 
-pub fn maximum(nums: Vec<i32>) -> i32 {
+pub fn maximum(nums: Vec<isize>) -> isize {
     let (mut s1, mut s2, mut b1, mut b2, mut b3) =
-        (i32::MAX, i32::MAX, i32::MIN, i32::MIN, i32::MIN);
+        (isize::MAX, isize::MAX, isize::MIN, isize::MIN, isize::MIN);
 
     for n in nums {
         if n < s1 {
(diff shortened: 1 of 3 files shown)


src/big_integer/maximum.rs
```diff
@@ -0,0 +1,33 @@
+/// Compute the multiplication set of the three largest numbers in an array
+pub fn maximum(mut nums: Vec<i32>) -> i32 {
+    nums.sort();
+    (nums[nums.len() - 1] * nums[nums.len() - 2] * nums[nums.len() - 3])
+        .max(nums[0] * nums[nums.len() - 1] * nums[nums.len() - 2])
+        .max(nums[0] * nums[1] * nums[nums.len() - 1])
+}
```
vil02, line 7: The complexity of this is $O(n\log n)$. I think it could be done in $O(n)$. Fell free to add some more implementations (in the same file).

src/big_integer/maximum.rs
vil02, line 1: This file definitely does not belong to the `big_integer` directory.

vil02 (changes requested):

src/big_integer/maximum.rs
mengfansheng-git: math？ or add array？

src/big_integer/maximum.rs
mengfansheng-git: I think Under src create a array under src would be a good choice.

src/big_integer/maximum.rs
mengfansheng-git: I think Under src create a array under src would be a good choice.

TruongNhanNguyen: Your PR is solving a problem from Leetcode, which often leads to unknowing algorithm categories. So we need to know the target of the algorithm (or data structure) so that we can clarify it into the best-suited category. In case, there are no suitable categories, we need to discuss whether it is necessary to create a new directory.

TruongNhanNguyen: - What do you think about change the type of elements from `i32` to `isize` (you may need to handle larger ranges of integer values or if your algorithm benefits from being able to handle platform-dependent integer sizes, which vary between 32-bit and 64-bit platforms)
- What do you think about renaming and adding some more tests, here is my suggestions (feel free to modify or apply on your own):
    ```rs
        test_maximum! {
            three_positive_numbers: ([1, 2, 3], 6),
            four_positive_numbers: ([1, 2, 3, 4], 24),
            four_numbers_with_zero: ([-1, 2, 0, 4], 0),
            three_negative_numbers: ([-1, -2, -3], -6),
            five_negative_numbers: ([-1, -2, -3, -4, -5], -6),
            negative_numbers_with_zero: ([-1, -2, 0, -3, -4, -5], 0),
            four_negative_numbers_with_zero: ([-1, -2, 0, -3, -4], 0),
            mixed_positive_and_negative_numbers1: ([-1, -2, -3, 4, 5], 30),
            mixed_positive_and_negative_numbers2: ([-1, 2, -3, -4, 5], 60),
            mixed_positive_and_negative_numbers3: ([-1, 2, -3, 4, 5], 40),
            single_negative_number_repeated: ([-1, -1, -1, -1, -1], -1),
            all_zeros: ([0, 0, 0, 0, 0], 0),
            two_positive_one_negative: ([1, 2, -3], -6),
            two_negative_one_positive: ([-1, -2, 3], 6),
            large_numbers: ([1000, 1000, 10000], 10000000000),
            mixture_with_large_and_small_numbers: ([1, -1, 2, -2, 1000000], 2000000),
        }
   ```

PR #730, opened 2024-06-01
Implement Rat in Maze Path Finder

TruongNhanNguyen:
## Description

See details in the implementation

## Type of change

## Checklist:


src/backtracking/rat_in_maze.rs
```diff
@@ -0,0 +1,225 @@
...
+    /// Constructs a new Maze instance.
+    /// # Arguments
+    ///
+    /// * `maze` - The maze represented as a 2D array where 1 represents a valid path and 0 represents a wall.
+    ///
+    /// # Returns
+    ///
+    /// A new Maze instance.
+    fn new(maze: Vec<Vec<usize>>) -> Self {
+        Maze { maze }
```
vil02, line 57: What if `maze` is not a proper maze (i.e. some of the _rows_ have different length)?

src/backtracking/rat_in_maze.rs
```diff
@@ -0,0 +1,225 @@
...
+    maze: Vec<Vec<usize>>,
+}
+
+impl Maze {
+    /// Represents possible moves in the maze.
+    const MOVES: [(isize, isize); 4] = [
+        (0, 1),  // Move right
+        (1, 0),  // Move down
+        (0, -1), // Move left
+        (-1, 0), // Move up
```
vil02, line 45:
```suggestion
        (0, 1),
        (1, 0),
        (0, -1),
        (-1, 0),
```

I have an impression that the _labels_ are mixed, for example `(0 ,1)` is not a _Move right_. Is that correct?

src/backtracking/rat_in_maze.rs
```diff
@@ -0,0 +1,225 @@
...
+            vec![0, 0, 0, 0, 0, 1, 1, 1],
+            vec![0, 0, 0, 0, 0, 0, 0, 1],
+            vec![0, 0, 0, 0, 0, 0, 0, 1],
+        ]),
+        test_find_path_in_maze_without_solution_4x4: 0, 0, &[
+            vec![1, 0, 0, 0],
+            vec![1, 1, 0, 0],
+            vec![0, 1, 1, 0],
+            vec![0, 0, 0, 1],
+        ], None::<Vec<Vec<usize>>>,
```
vil02, line 223:
```suggestion
        ], None::<Vec<Vec<usize>>>,
        no_solution_possible_loops: 0, 0, &[
            vec![1, 1, 1, 1],
            vec![1, 1, 1, 1],
            vec![1, 1, 1, 0],
            vec![1, 1, 0, 1],
        ], None::<Vec<Vec<usize>>>,
        start_outside: 5, 0, &[
            vec![1, 0, 0, 0],
            vec![1, 1, 0, 0],
            vec![0, 1, 1, 0],
            vec![0, 0, 1, 1],
        ], None::<Vec<Vec<usize>>>,
```

vil02 (changes requested):

src/backtracking/rat_in_maze.rs
TruongNhanNguyen: Yes, I make the wrong comments there.

src/backtracking/rat_in_maze.rs
TruongNhanNguyen: To handle the case where the `maze` is not a proper maze (i.e., some of the rows have different lengths), we could add validation logic to ensure that the maze is rectangular before proceeding with the algorithm. If the `maze` is not rectangular, the function should return an error or handle the case appropriately.

But, I think adding these validation logics is all unnecessary because the `maze` is always supposed to be an `NxN` 2D array.

src/backtracking/rat_in_maze.rs
vil02: I would be careful with _up_ and _down_. I think that the variable name `MOVES` is descriptive enough.

src/backtracking/rat_in_maze.rs
vil02: What if someone will call your function with input like `&[vec![1], vec![1, 1], vec![1, 1, 1], vec![1, 1, 1, 1]]`?

By the way, why not to use `Vec<Vec<bool>>`?

src/backtracking/rat_in_maze.rs
TruongNhanNguyen: I will remove the redundant comments

src/backtracking/rat_in_maze.rs
TruongNhanNguyen: - I will modify the implementation to handle the cases of improper mazes.
- I used `Vec<Vec<usize>>` for explicit representation of the maze (0 stands for the wall, 1 stands for the path). I know that `Vec<Vec<bool>>` typically occupies less memory than `usize` (or `isize`). Each `bool` in Rust is represented as a single byte, while `usize` generally takes up more space (typically 4 or 8 bytes depending on the platform).
- In most cases for representing a maze where cells are either passable or blocked, using Vec<Vec<bool>> is a straightforward and memory-efficient choice. It aligns well with the problem domain (maze traversal) and improves code clarity. 
- I will change the implementation to use `Vec<Vec<bool>>` as maze representation.

TruongNhanNguyen: If memory-efficient is really more important than maze representation. I would chang the implementation of maze to `Vec<Vec<bool>>`. What do you think @vil02.

vil02: Memory efficiency is one thing. The other thing is that if you will use `Vec<Vec<bool>>` you have only _two states_. You do not need to worry about the inputs like `&[vec![1, 5], vec![0, 1], vec![0, 1]]`.

src/backtracking/rat_in_maze.rs
```diff
@@ -0,0 +1,250 @@
...
+            vec![1, 1, 0, 0],
+            vec![0, 1, 1, 0],
+            vec![0, 0, 1, 1]
+        ]),
+        unproper_maze_no_solution_possible_loops: 0, 0, &[
+            vec![1, 1, 1, 1],
+            vec![1, 1],
+            vec![1, 1, 0],
+            vec![0, 0, 1],
+        ], None::<Vec<Vec<usize>>>,
```
vil02, line 248: One more thing: could you please add a test case with input not being a square, for example `2x3` like input?

TruongNhanNguyen: I will try

src/backtracking/rat_in_maze.rs
```diff
@@ -0,0 +1,267 @@
...
+        Maze { maze }
+    }
+
+    /// Returns the width of the maze.
+    ///
+    /// # Returns
+    ///
+    /// The width of the maze.
+    fn width(&self) -> usize {
+        self.maze.iter().map(|row| row.len()).max().unwrap_or(0)
```
vil02, line 62: This is quite complex. Why don't you just return some error if the input is not a rectangle?

Or, since this project [depends on `nalgebra`](https://github.com/TheAlgorithms/Rust/blob/864ef42a427cb1f841d85c60b96c01c63154e54c/Cargo.toml#L13), you could use some matrix type provided there.

I prefer the first approach.

TruongNhanNguyen: What do you mean by returning some error if the input is not a rectangle? Currently, the implementation handles the maze in both rectangle and non-rectangle representation. This function returns the width of the maze, which is equal to the largest row in the maze.

vil02: The signature of `find_path_in_maze` could look like:
```rust
pub fn find_path_in_maze(
    maze: &[Vec<bool>],
    start_x: usize,
    start_y: usize,
) -> Result<Option<Vec<Vec<bool>>>, SomeInvalidInputError>
```

TruongNhanNguyen: Could you review the update, @vil02. I have adjusted the implementation of the function `find_path_in_maze` to handle exception cases: empty maze, start out-of-bound on the maze, and non-rectangle maze. I also added some edge tests to cover various scenarios of maze pathfinding.

src/backtracking/rat_in_maze.rs
```diff
@@ -0,0 +1,304 @@
...
+        ], Ok(None::<Vec<Vec<bool>>>),
+        maze_with_no_walls: 0, 0, &[
+            vec![true, true, true],
+            vec![true, true, true],
+            vec![true, true, true],
+        ], Ok(Some(vec![
+            vec![true, true, true],
+            vec![false, false, true],
+            vec![false, false, true],
+        ])),
```
vil02, line 302:
```suggestion
        ])),
        trivial_maze: 0, 0, &[
            vec![true],
        ], Ok(Some(vec![
            vec![true],
        ])),
        maze_with_going_back: 0, 0, &[
            vec![true,  true,  true,  true, true,   true],
            vec![false, false, false, true, false,  true],
            vec![true,  true,  true,  true,  false, false],
            vec![true,  false, false, false, false, false],
            vec![true,  false, false, false, true, true],
            vec![true,  false, true,  true,  true,  false],
            vec![true,  false, true , false, true,  false],
            vec![true,  true,  true,  false, true,  true],
        ], Ok(Some(vec![
            vec![true,  true,  true,  true, false,  false],
            vec![false, false, false, true, false,  false],
            vec![true,  true,  true,  true,  false, false],
            vec![true,  false, false, false, false, false],
            vec![true,  false, false, false, false, false],
            vec![true,  false, true,  true,  true,  false],
            vecargoc![true,  false, true , false, true,  false],
            vec![true,  true,  true,  false, true,  true],
        ])),
```

vil02, line 302: Do you have some good idea to _document_ why your implementation returns this solution and not the other one? I guess it would be something like: If multiple solutions exists, the function returns the first _successful path_ it discovers, based on the predefined order of considered moves, defined in `MOVES` .

That would be an overkill to handle all of the solutions.

vil02 (changes requested):

src/backtracking/rat_in_maze.rs
TruongNhanNguyen: You're right. Handle the case that return all possible path is overkill. I have commented what is solution matrix. Solution matrix is contain one possible path, based on the order of predefined moves.

TruongNhanNguyen: I will add document explain it and the trivial test is current test with single cell

vil02 (approved):

PR #731, opened 2024-06-01
Implement Hamiltonian Cycle Finder

TruongNhanNguyen:
# Pull Request Template

## Description

See details in the implementation

## Type of change

## Checklist:


TruongNhanNguyen:
Can you have a look on this PR (and other PRs in the backtracking category) and review them @vil02.

TruongNhanNguyen:
Can you review this PR, @vil02?

src/backtracking/hamiltonian_cycle.rs
```diff
@@ -0,0 +1,180 @@
+//! This module provides functionality to find a Hamiltonian cycle in graph.
+//! Source: [Wikipedia](https://en.wikipedia.org/wiki/Hamiltonian_path_problem)
+
+/// Represents a graph with an adjacency matrix.
+struct Graph {
+    /// The adjacency matrix representing the graph.
+    adjacency_matrix: Vec<Vec<u8>>,
```
vil02, line 7:
```suggestion
    adjacency_matrix: Vec<Vec<bool>>,
```
And as [before](https://github.com/TheAlgorithms/Rust/pull/730#discussion_r1639932353): what if the input will not be _rectangular_? Maybe it would make sense to introduce a `struct` like `AdjacencyMatrix`.

src/backtracking/hamiltonian_cycle.rs
```diff
@@ -0,0 +1,180 @@
...
+    /// The number of vertices in the graph.
+    vertices: usize,
```
vil02, line 9:
```suggestion
```
This is redundant. It could be read from `adjacency_matrix`.

src/backtracking/hamiltonian_cycle.rs
```diff
@@ -0,0 +1,180 @@
...
+        } else {
+            None
+        }
+    }
+}
+
+/// Finds a Hamiltonian cycle in a given graph represented by an adjacency matrix, if one exists, starting from a specified vertex
+pub fn find_hamiltonian_cycle(
+    adjacency_matrix: Vec<Vec<u8>>,
+    start_vertex: usize,
```
vil02, line 105: What if `start_vertex` is _out of range_?

src/backtracking/hamiltonian_cycle.rs
```diff
@@ -0,0 +1,180 @@
...
+            return false;
+        }
+
+        // Check if the vertex has already been included in the path
+        !path[..pos].contains(&v)
+    }
+
+    /// Utility function for finding a Hamiltonian cycle recursively.
+    ///
+    /// This function is called recursively by `find_hamiltonian_cycle`.
```
vil02, line 51: This sentence does not make sense.

src/backtracking/hamiltonian_cycle.rs
```diff
@@ -0,0 +1,180 @@
...
+
+    /// Finds a Hamiltonian cycle in the graph, if one exists, starting from the specified vertex.
+    ///
+    /// A Hamiltonian cycle is a cycle that visits every vertex exactly once
+    /// and returns to the starting vertex.
+    ///
+    /// Returns `Some(path)` if a Hamiltonian cycle is found, where `path` is a vector
+    /// containing the indices of vertices in the cycle, starting and ending with the same vertex.
+    ///
+    /// Returns `None` if no Hamiltonian cycle exists in the graph.
```
vil02, line 88: Please add a comment regarding non-unique cycles.

src/backtracking/hamiltonian_cycle.rs
```diff
@@ -0,0 +1,180 @@
...
+    use super::*;
+
+    macro_rules! hamiltonian_cycle_tests {
+        ($($name:ident: $test_case:expr,)*) => {
+            $(
+                #[test]
+                fn $name() {
+                    let (adjacency_matrix, start_vertex, expected) = $test_case;
+                    let result = find_hamiltonian_cycle(adjacency_matrix, start_vertex);
+                    assert_eq!(result, expected);
```
vil02, line 122: What do you think about extending this test by _rotating_ the `expected` and calling `find_hamiltonian_cycle` with `start_vertex` being the the first element of `expected`?

vil02 (changes requested):

src/backtracking/hamiltonian_cycle.rs
TruongNhanNguyen: It should be removed.

src/backtracking/hamiltonian_cycle.rs
TruongNhanNguyen: I will add custom error types named `AdjMatError` with 3 variants: `EmptyMat`, `StartOutOfBound` and `ImproperMat` to handle exceptions.

src/backtracking/hamiltonian_cycle.rs
TruongNhanNguyen: I will rewrite the docs with relevant comments.

src/backtracking/hamiltonian_cycle.rs
TruongNhanNguyen: Sure, I will do it.

src/backtracking/hamiltonian_cycle.rs
TruongNhanNguyen: It's a good idea.

TruongNhanNguyen: I think there is a simple way to solve the problem. We can provide the test inputs containing an adjacency matrix as the first argument, followed by a tuple of start vertices and their expected cycles as the second and third arguments. What do you think about this, @vil02? Note that: it maybe not be suited for large graphs.

vil02: That might be too much. I will _mark is as resolved_ for now.

src/backtracking/hamiltonian_cycle.rs
```diff
@@ -0,0 +1,238 @@
...
+
+    // Validate maze representation (if necessary)
+    if adjacency_matrix.iter().any(|row| row.len() != vertices) {
+        return Err(AdjMatError::ImproperMat);
+    }
+
+    // Validate start position
+    if start_vertex >= vertices {
+        return Err(AdjMatError::StartOutOfBound);
+    }
```
vil02, line 138: What do you think about moving this _logic_ into the `new` method in `Graph`? That would require the change of its signature to:
```rust
fn new(adjacency_matrix: Vec<Vec<bool>>) -> Resut<Self, AdjMatError>
```

src/backtracking/hamiltonian_cycle.rs
```diff
@@ -0,0 +1,238 @@
...
+        }
+    }
+}
+
+/// Finds a Hamiltonian cycle in a given graph represented by an adjacency matrix, if one exists, starting from a specified vertex
+pub fn find_hamiltonian_cycle(
+    adjacency_matrix: Vec<Vec<bool>>,
+    start_vertex: usize,
+) -> Result<Option<Vec<usize>>, AdjMatError> {
+    let vertices = adjacency_matrix.len();
```
vil02, line 124: `Graph` should have a method returning the number of vertices.

vil02 (changes requested):

src/backtracking/hamiltonian_cycle.rs
TruongNhanNguyen: I have moved this logic into appropriate methods and changed the method signatures.

vil02:
Failing CI will be fixed by #754.

src/backtracking/hamiltonian_cycle.rs
```diff
@@ -0,0 +1,248 @@
...
+/// Represents potential errors when working with an adjacency matrix.
+#[derive(Debug, PartialEq, Eq)]
+pub enum AdjMatError {
+    /// Indicates that the adjacency matrix is empty.
+    EmptyMat,
+    /// Indicates that the adjacency matrix is not square.
+    ImproperMat,
+    /// Indicates that the starting vertex is out of bounds.
+    StartOutOfBound,
+}
```
vil02, line 13: I am thinking about naming here. First of all, maybe `AdjacencyMatrixError` would be better. Secondly, is `StartOutOfBound` a problem of the adjacency matrix or the starting position. Maybe it should be simply `FindHamiltonianCycleError` or something less generic.

src/backtracking/hamiltonian_cycle.rs
```diff
@@ -0,0 +1,248 @@
...
+    ///
+    /// `true` if it is safe to include `v` in the path, `false` otherwise.
+    fn is_safe(&self, v: usize, path: &[usize], pos: usize) -> bool {
+        // Check if the current vertex and the last vertex in the path are adjacent.
+        if !self.adjacency_matrix[path[pos - 1]][v] {
+            return false;
+        }
+
+        // Check if the vertex has already been included in the path.
+        !path[..pos].contains(&v)
```
vil02, line 73: This is quite inefficient - $O(n)$. Equivalent computation could be done in constant time, when there is some `visited` variable being a hash set.

src/backtracking/hamiltonian_cycle.rs
```diff
@@ -0,0 +1,248 @@
...
+    }
+}
+
+/// Attempts to find a Hamiltonian cycle in a graph represented by an adjacency matrix, starting from a specified vertex.
+pub fn find_hamiltonian_cycle(
+    adjacency_matrix: Vec<Vec<bool>>,
+    start_vertex: usize,
+) -> Result<Option<Vec<usize>>, AdjMatError> {
+    let graph = Graph::new(adjacency_matrix)?;
+    graph.find_hamiltonian_cycle(start_vertex)
```
vil02, line 152:
```suggestion
    Graph::new(adjacency_matrix)?.find_hamiltonian_cycle(start_vertex)
```

src/backtracking/hamiltonian_cycle.rs
```diff
@@ -0,0 +1,248 @@
...
+            vec![
+                vec![false, true, false, true, false],
+                vec![true, false, true, true, false],
+                vec![false, true, false, false, false],
+                vec![true, true, false, false, true],
+                vec![false, false, false, true, false],
+            ],
+            0,
+            Ok(None::<Vec<usize>>)
+        ),
```
vil02, line 222: Does this algorithm work for directed graphs? If I am not mistaking, all of the provided adjacency matrices are symmetric, so the represented graph is undirected.

vil02 (changes requested):

src/backtracking/hamiltonian_cycle.rs
TruongNhanNguyen: The `FindHamiltonianCycleError` is a better name for all these errors. The error variants related to the adjacency matrix will be renamed to `EmptyAdjMat` and `ImproperAdjMat` for better.

src/backtracking/hamiltonian_cycle.rs
TruongNhanNguyen: I will refactor the logic to make it more efficient.

src/backtracking/hamiltonian_cycle.rs
TruongNhanNguyen: Currently, the implementation only supports undirected graphs.

TruongNhanNguyen: Test for directed graph passed. You can see these tests in the implementation.

src/backtracking/hamiltonian_cycle.rs
```diff
@@ -0,0 +1,298 @@
...
+
+use std::collections::HashSet;
+
+/// Represents potential errors when finding hamiltonian cycle on an adjacency matrix.
+#[derive(Debug, PartialEq, Eq)]
+pub enum FindHamiltonianCycleError {
+    /// Indicates that the adjacency matrix is empty.
+    EmptyAdjMat,
+    /// Indicates that the adjacency matrix is not square.
+    ImproperAdjMat,
```
vil02, line 12: Why not to use full words here?

src/backtracking/hamiltonian_cycle.rs
```diff
@@ -0,0 +1,298 @@
...
+    ///
+    /// * `v` - The index of the vertex being considered.
+    /// * `visited` - A reference to the set of visited vertices.
+    /// * `path` - A reference to the current path being explored.
+    /// * `pos` - The position of the current vertex being considered.
+    ///
+    /// # Returns
+    ///
+    /// `true` if it is safe to include `v` in the path, `false` otherwise.
+    fn is_safe(&self, v: usize, visited: &HashSet<usize>, path: &[usize], pos: usize) -> bool {
```
vil02, line 69: Using `HashSet` here might be an overkill. Since the vertices are numbered from zero, one could consider using `Vec<bool>`, with some _possible_ performance gain.

src/backtracking/hamiltonian_cycle.rs
```diff
@@ -0,0 +1,298 @@
...
+        &self,
+        start_vertex: usize,
+    ) -> Result<Option<Vec<usize>>, FindHamiltonianCycleError> {
+        // Validate the start vertex.
+        if start_vertex >= self.num_vertices() {
+            return Err(FindHamiltonianCycleError::StartOutOfBound);
+        }
+
+        // Initialize the path.
+        let mut path = vec![usize::MAX; self.adjacency_matrix.len()];
```
vil02, line 143: Thing to consider: why not to represent `path` as `Vec<Option<usize>>`. `None`, would represent the _undefined_ node.

Furthermore:
```suggestion
        let mut path = vec![usize::MAX; self.num_vertices()];
```

src/backtracking/hamiltonian_cycle.rs
```diff
@@ -0,0 +1,298 @@
...
+                vec![false, true, false, true, false, false],
+                vec![false, false, true, false, true, false],
+                vec![false, false, false, true, false, false],
+                vec![false, true, false, false, true, false],
+                vec![false, false, true, false, false, true],
+                vec![true, false, false, false, false, false],
+            ],
+            0,
+            Ok(Some(vec![0, 1, 2, 3, 4, 5, 0]))
+        ),
```
vil02, line 296:
```suggestion
        ),
        single_node_self_loop: (
            vec![
                vec![true],
            ],
            0,
            Ok(Some(vec![0, 0]))
        ),
        single_node: (
            vec![
                vec![false],
            ],
            0,
            Ok(None)
        ),
```

vil02 (changes requested):

src/backtracking/hamiltonian_cycle.rs
TruongNhanNguyen: It's a good option to implement. Using a `Vec<bool>` instead of a `HashSet<usize>` can indeed provide a performance boost since it allows for constant-time access and updates, and it typically consumes less memory.

src/backtracking/hamiltonian_cycle.rs
TruongNhanNguyen: These are abbreviations.

src/backtracking/hamiltonian_cycle.rs
TruongNhanNguyen: It's a good suggestion, I will consider it.

src/backtracking/hamiltonian_cycle.rs
TruongNhanNguyen: I will use full word as you suggested.

vil02 (approved):

PR #734, opened 2024-06-03
Implementation of Negative Log Likelihood Loss Function

jkauerl:
# Pull Request Template

## Description

PR including the implementation for the Negative Log Likelihood Loss Function of this issue #559 . It creates a new file called `negative_log_likelihood.rs` with the function called `neg_log_likelihood`. This implementation is for binary classifications, but can be expanded to higher dimensions.

## Type of change

## Checklist:


src/machine_learning/loss_function/negative_log_likelihood.rs
```diff
@@ -0,0 +1,38 @@
...
+// elements.
+
+pub fn neg_log_likelihood(y_true: &[f64], y_pred: &[f64]) -> f64 {
+    let mut total_loss: f64 = 0.0;
+    for (p, a) in y_pred.iter().zip(y_true.iter()) {
+        let loss: f64 = -a * p.ln() - (1.0 - a) * (1.0 - p).ln();
+        total_loss += loss;
+    }
+    total_loss / (y_pred.len() as f64)
+}
```
vil02, line 23: There are few edge cases which could be handled:
- what if `y_true` has different length than `y_pred`?
- what if `y_pred` is empty?
- what if you cannot compute the logarithms? I think you assume that your inputs are in $[0, 1]$ - please add an explicit check of that.

vil02 (changes requested):

src/machine_learning/loss_function/negative_log_likelihood.rs
jkauerl: Thanks for the quick response. Will glady add those different checks to the different inputs.

jkauerl: > There are few edge cases which could be handled:
> 
> * what if `y_true` has different length than `y_pred`?
> * what if `y_pred` is empty?
> * what if you cannot compute the logarithms? I think you assume that your inputs are in (0,1) - please add an explicit check of that.

Hello, just added the checks for the input values, and some test cases.

src/machine_learning/loss_function/negative_log_likelihood.rs
```diff
@@ -0,0 +1,77 @@
...
+        let predicted_values: Vec<f64> = vec![1.1, 0.1, 0.8];
+        assert_eq!(neg_log_likelihood(&actual_values, &predicted_values), None);
+    }
+
+    #[test]
+    fn test_neg_log_likelihood_empty_prediction() {
+        let actual_values: Vec<f64> = vec![1.0, 0.0, 1.0];
+        let predicted_values: Vec<f64> = vec![];
+        assert_eq!(neg_log_likelihood(&actual_values, &predicted_values), None);
+    }
```
vil02, line 76:
```suggestion
    macro_rules! test_with_wrong_inputs {
        ($($name:ident: $inputs:expr,)*) => {
            $(
                #[test]
                fn $name() {
                    let (values_a, values_b) = $inputs;
                    assert!(neg_log_likelihood(&values_a, &values_b).is_none());
                    assert!(neg_log_likelihood(&values_b, &values_a).is_none());
                }
            )*
        }
    }

    test_with_wrong_inputs! {
        different_length: (vec![0.9, 0.0, 0.8], vec![0.9, 0.1]),
        different_length_one_empty: (vec![], vec![0.9, 0.1]),
        value_greater_than_1: (vec![1.1, 0.0, 0.8], vec![0.1, 0.2, 0.3]),
        value_greater_smaller_than_0: (vec![0.9, 0.0, -0.1], vec![0.1, 0.2, 0.3]),
        empty_input: (vec![], vec![]),
    }
```

src/machine_learning/loss_function/negative_log_likelihood.rs
```diff
@@ -0,0 +1,77 @@
...
+// `y_pred`, the Negative Log Likelihood loss is calculated as:
+//
+// - loss = `-y_true * log(y_pred) - (1 - y_true) * log(1 - y_pred)`.
+//
+// It returns the average loss by dividing the `total_loss` by total no. of
+// elements.
+
+pub fn neg_log_likelihood(y_true: &[f64], y_pred: &[f64]) -> Option<f64> {
+    // Checks if the length of the actual and predicted values are equal
+    if y_true.len() != y_pred.len() || y_pred.is_empty() || y_true.is_empty() {
```
vil02, line 18: What do you think about simplifying this condition?

```suggestion
    if y_true.len() != y_pred.len() || y_pred.is_empty() {
```
If I am not wrong, it is equivalent.

src/machine_learning/loss_function/negative_log_likelihood.rs
```diff
@@ -0,0 +1,77 @@
...
+        return None;
+    }
+    // Checks values are between 0 and 1
+    for (p, a) in y_pred.iter().zip(y_true.iter()) {
+        if *p < 0.0 || *p > 1.0 || *a < 0.0 || *a > 1.0 {
+            return None;
+        }
+    }
```
vil02, line 26:
```suggestion
    if !are_all_values_in_range(y_true) || !are_all_values_in_range(y_pred) {
        return None;
    }
```
where
```rust
fn are_all_values_in_range(values: &[f64]) -> bool {
    values.iter().all(|&x| (0.0..=1.0).contains(&x))
}
```

src/machine_learning/loss_function/negative_log_likelihood.rs
```diff
@@ -0,0 +1,77 @@
...
+//
+// ## Formula
+//
+// For a pair of actual and predicted values, represented as vectors `y_true` and
+// `y_pred`, the Negative Log Likelihood loss is calculated as:
+//
+// - loss = `-y_true * log(y_pred) - (1 - y_true) * log(1 - y_pred)`.
+//
+// It returns the average loss by dividing the `total_loss` by total no. of
+// elements.
```
vil02, line 14: Please add some external link.

src/machine_learning/loss_function/negative_log_likelihood.rs
```diff
@@ -0,0 +1,77 @@
...
+
+    #[test]
+    fn test_neg_log_likelihood() {
+        let actual_values: Vec<f64> = vec![1.0, 0.0, 1.0];
+        let predicted_values: Vec<f64> = vec![0.9, 0.1, 0.8];
+        assert_eq!(
+            neg_log_likelihood(&actual_values, &predicted_values),
+            Some(0.14462152754328741)
+        );
+    }
```
vil02, line 48: Please add some more examples. Use similar approach as suggested below.

vil02 (changes requested):

vil02:
@jkauerl are you still working on that?

jkauerl:
> @jkauerl are you still working on that?

Yes, I implemented most of the suggestions, and do not know what else is missing.

src/machine_learning/loss_function/negative_log_likelihood.rs
vil02: @jkauerl please have a look here.

src/machine_learning/loss_function/negative_log_likelihood.rs
vil02: @jkauerl: and here.

src/machine_learning/loss_function/negative_log_likelihood.rs
vil02: @jkauerl: and here.

src/machine_learning/loss_function/negative_log_likelihood.rs
vil02: @jkauerl: and here.

src/machine_learning/loss_function/negative_log_likelihood.rs
vil02: @jkauerl: and here.

src/machine_learning/loss_function/negative_log_likelihood.rs
jkauerl: I tried implementing the macro and when I did the testing, it took longer than the allowed time.

src/machine_learning/loss_function/negative_log_likelihood.rs
jkauerl: I was thinking of leaving the condition because maybe some could input vectors with different sizes.

src/machine_learning/loss_function/negative_log_likelihood.rs
jkauerl: Added more examples.

src/machine_learning/loss_function/negative_log_likelihood.rs
vil02: Could you elaborate more on that? This sounds really strange. There should be no impact on the performance.

src/machine_learning/loss_function/negative_log_likelihood.rs
vil02: > I was thinking of leaving the condition because maybe some could input vectors with different sizes.

They cannot be, because of `y_true.len() != y_pred.len()`.

src/machine_learning/loss_function/negative_log_likelihood.rs
vil02: > Added more examples.

Where are they? Your last commit, 65671b79add3ba11196bf690df47cfe2c50e7743, is from 19.06.2024.

src/machine_learning/loss_function/negative_log_likelihood.rs
```diff
@@ -0,0 +1,95 @@
...
+// `y_pred`, the Negative Log Likelihood loss is calculated as:
+//
+// - loss = `-y_true * log(y_pred) - (1 - y_true) * log(1 - y_pred)`.
+//
+// It returns the average loss by dividing the `total_loss` by total no. of
+// elements.
+//
+// https://towardsdatascience.com/cross-entropy-negative-log-likelihood-and-all-that-jazz-47a95bd2e81
+
+pub fn neg_log_likelihood(y_true: &[f64], y_pred: &[f64]) -> Option<f64> {
```
vil02, line 18: What do you think this:
```suggestion
pub fn neg_log_likelihood(y_true: &[f64], y_pred: &[f64]) -> Result<f64, NegativeLogLikelihoodLossError> {
```
with
```rust
#[derive(Debug, PartialEq, Eq)]
pub enum NegativeLogLikelihoodLossError {
    InputsHaveDifferentLength,
    EmptyInputs,
}
```

The `Option` should be used in situations when _solution does not exist for some reasonable inputs_. Here the inputs are _obviously_ incorrect.

src/machine_learning/loss_function/negative_log_likelihood.rs
jkauerl: Oh right, just updated it.

src/machine_learning/loss_function/negative_log_likelihood.rs
jkauerl: This is implemented.

src/machine_learning/loss_function/negative_log_likelihood.rs
jkauerl: This is also done, but it is below this code.

src/machine_learning/loss_function/negative_log_likelihood.rs
jkauerl: Although, I think is important to include the condition because there can be 2 errors associated. One related to an empty vector, and another with 2 vectors with different sizes. So then it would be neccessary to include both conditions?

src/machine_learning/loss_function/negative_log_likelihood.rs
jkauerl: Makes sense for me and just implemented it. But I added one more error which is for invalid values. Is that right or should I change it back?

Like this:
``` python
#[derive(Debug, PartialEq, Eq)]
pub enum NegativeLogLikelihoodLossError {
    InputsHaveDifferentLength,
    EmptyInputs,
    InvalidValues,
}
```

src/machine_learning/loss_function/negative_log_likelihood.rs
jkauerl: I implemented 7 tests checking for correct calculation, invalid length, invalid values, empty values and negative values. Is there any more tests missing?

src/machine_learning/loss_function/negative_log_likelihood.rs
jkauerl: Oh, if there should not be an impact on performance then it might be the computer I am using.

src/machine_learning/loss_function/negative_log_likelihood.rs
```diff
@@ -0,0 +1,128 @@
...
+    y_pred: &[f64],
+) -> Result<f64, NegativeLogLikelihoodLossError> {
+    // Checks if the inputs are empty
+    if y_pred.is_empty() || y_true.is_empty() {
+        return Err(NegativeLogLikelihoodLossError::EmptyInputs);
+    }
+    // Checks if the length of the actual and predicted values are equal
+    if y_true.len() != y_pred.len() {
+        return Err(NegativeLogLikelihoodLossError::InputsHaveDifferentLength);
+    }
```
vil02, line 29:
```suggestion
    if y_true.len() != y_pred.len() {
        return Err(NegativeLogLikelihoodLossError::InputsHaveDifferentLength);
    }

    if y_pred.is_empty() {
        return Err(NegativeLogLikelihoodLossError::EmptyInputs);
    }
```
Note that at the second `if`, it is sure that the inputs have same lengths. Thanks to this change, the code has lower cyclomatic complexity.

src/machine_learning/loss_function/negative_log_likelihood.rs
```diff
@@ -0,0 +1,128 @@
...
+
+    #[test]
+    fn test_neg_log_likelihood_negative_values1() {
+        let actual_values: Vec<f64> = vec![1.0, 0.0, 1.0];
+        let predicted_values: Vec<f64> = vec![0.9, 0.1, -0.8];
+        assert_eq!(
+            neg_log_likelihood(&actual_values, &predicted_values),
+            Err(NegativeLogLikelihoodLossError::InvalidValues)
+        );
+    }
```
vil02, line 127: I would try again:

```suggestion
    macro_rules! test_with_wrong_inputs {
        ($($name:ident: $inputs:expr,)*) => {
            $(
                #[test]
                fn $name() {
                    let (values_a, values_b, expected_error) = $inputs;
                    assert_eq!(neg_log_likelihood(&values_a, &values_b), expected_error);
                    assert_eq!(neg_log_likelihood(&values_b, &values_a), expected_error);
                }
            )*
        }
    }

    test_with_wrong_inputs! {
        different_length: (vec![0.9, 0.0, 0.8], vec![0.9, 0.1], Err(NegativeLogLikelihoodLossError::InputsHaveDifferentLength)),
        different_length_one_empty: (vec![], vec![0.9, 0.1], Err(NegativeLogLikelihoodLossError::InputsHaveDifferentLength)),
        value_greater_than_1: (vec![1.1, 0.0, 0.8], vec![0.1, 0.2, 0.3], Err(NegativeLogLikelihoodLossError::InvalidValues)),
        value_greater_smaller_than_0: (vec![0.9, 0.0, -0.1], vec![0.1, 0.2, 0.3], Err(NegativeLogLikelihoodLossError::InvalidValues)),
        empty_input: (vec![], vec![], Err(NegativeLogLikelihoodLossError::EmptyInputs)),
    }
```
Advantages:
- the code is shorter,
- more _paths_ are tested (6 vs. 8).

Please note, that this change required changing the order of `if`'s.

src/machine_learning/loss_function/negative_log_likelihood.rs
```diff
@@ -0,0 +1,128 @@
...
+
+    #[test]
+    fn test_neg_log_likelihood_invalid_length0() {
+        let actual_values: Vec<f64> = vec![1.0, 0.0, 1.0];
+        let predicted_values: Vec<f64> = vec![0.9, 0.1];
+        assert_eq!(
+            neg_log_likelihood(&actual_values, &predicted_values),
+            Err(NegativeLogLikelihoodLossError::InputsHaveDifferentLength)
+        );
+    }
```
vil02, line 77:
```suggestion
    macro_rules! test_neg_log_likelihood {
        ($($name:ident: $inputs:expr,)*) => {
            $(
                #[test]
                fn $name() {
                    let (actual_values, predicted_values, expected) = $inputs;
                    assert_eq!(neg_log_likelihood(&actual_values, &predicted_values).unwrap(), expected);
                }
            )*
        }
    }

    test_neg_log_likelihood! {
        set_0: (vec![1.0, 0.0, 1.0], vec![0.9, 0.1, 0.8], 0.14462152754328741),
        // one should add more here
    }
```

vil02 (changes requested):

src/machine_learning/loss_function/negative_log_likelihood.rs
```diff
@@ -0,0 +1,98 @@
...
+//
+// For a pair of actual and predicted values, represented as vectors `y_true` and
+// `y_pred`, the Negative Log Likelihood loss is calculated as:
+//
+// - loss = `-y_true * log(y_pred) - (1 - y_true) * log(1 - y_pred)`.
+//
+// It returns the average loss by dividing the `total_loss` by total no. of
+// elements.
+//
+// https://towardsdatascience.com/cross-entropy-negative-log-likelihood-and-all-that-jazz-47a95bd2e81
```
vil02, line 16: This page focuses on the _binary case_. Do you know some source, which has exactly the same formula as you wrote above?

src/machine_learning/loss_function/negative_log_likelihood.rs
```diff
@@ -0,0 +1,101 @@
...
+                }
+            )*
+        }
+    }
+
+    test_neg_log_likelihood! {
+        set_0: (vec![1.0, 0.0, 1.0], vec![0.9, 0.1, 0.8], 0.14462152754328741),
+        set_1: (vec![1.0, 0.0, 1.0], vec![0.1, 0.2, 0.3], 1.2432338162113972),
+        set_2: (vec![0.0, 1.0, 0.0], vec![0.1, 0.2, 0.3], 0.6904911240102196),
+
```
vil02, line 98:
```suggestion
```

vil02 (approved):

PR #737, opened 2024-06-06
Implement Graph Coloring using Backtracking

TruongNhanNguyen:
## Description

See in the implementation.

## Type of change

## Checklist:

### DIFF
--- a/src/backtracking/graph_coloring.rs
+++ b/src/backtracking/graph_coloring.rs
@@ -1,30 +1,41 @@
-//! This module provides functionality for generating all possible colorings of a graph
+//! This module provides functionality for generating all possible colorings of a undirected (or directed) graph
 //! given a certain number of colors. It includes the GraphColoring struct and methods
 //! for validating color assignments and finding all valid colorings.
+
+/// Represents potential errors when coloring on an adjacency matrix.
+#[derive(Debug, PartialEq, Eq)]
+pub enum GraphColoringError {
+    // Indicates that the adjacency matrix is empty
+    EmptyGraph,
+    // Indicates that the adjacency matrix is not squared
+    NonSquareMatrix,
+}
 
 /// Generates all possible valid colorings of a graph.
 ///
 /// # Arguments
 ///
-/// * `adj_matrix` - A 2D vector representing the adjacency matrix of the graph.
+/// * `adjacency_matrix` - A 2D vector representing the adjacency matrix of the graph.
 /// * `num_colors` - The number of colors available for coloring the graph.
 ///
 /// # Returns
 ///
-/// * An `Option` containing a vector of solutions. Each solution is a vector of color assignments for the vertices.
+/// * A `Result` containing an `Option` with a vector of solutions or a `GraphColoringError` if
+/// there is an issue with the matrix.
 pub fn generate_colorings(
-    adj_matrix: Vec<Vec<bool>>,
+    adjacency_matrix: Vec<Vec<bool>>,
     num_colors: usize,
-) -> Option<Vec<Vec<usize>>> {
-    let mut graph_coloring = GraphColoring::new(adj_matrix, num_colors);
-    graph_coloring.find_solutions()
+) -> Result<Option<Vec<Vec<usize>>>, GraphColoringError> {
+    GraphColoring::new(adjacency_matrix)?.find_solutions(num_colors)
 }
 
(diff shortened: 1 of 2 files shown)


TruongNhanNguyen:
Can you help a look on this PR, @vil02?

src/backtracking/graph_coloring.rs
```diff
@@ -0,0 +1,264 @@
...
+    adj_matrix: Vec<Vec<bool>>,
+    num_colors: usize,
+) -> Option<Vec<Vec<usize>>> {
+    let mut graph_coloring = GraphColoring::new(adj_matrix, num_colors);
+    graph_coloring.find_solutions()
+}
+
+/// A struct representing a graph coloring problem.
+struct GraphColoring {
+    adj_matrix: Vec<Vec<bool>>,
```
vil02, line 25: Is there a check ensuring that the `adj_matrix` (please rename) is a square?

Furthermore, in this case I guess you really want to have a directed graph (i.e. symmetric adjacency matrix).

src/backtracking/graph_coloring.rs
```diff
@@ -0,0 +1,264 @@
...
+    num_colors: usize,
```
vil02, line 26: This is redundant.
```suggestion
```

vil02 (changes requested):

src/backtracking/graph_coloring.rs
TruongNhanNguyen: I will add an assertion to check the adjacency matrix for squares. I also adjust the implementation to work with directed graphs.

src/backtracking/graph_coloring.rs
TruongNhanNguyen: I will remove it.

src/backtracking/graph_coloring.rs
TruongNhanNguyen: We can create a custom error type to handle the cases where the adjacency matrix is empty or not squared.

TruongNhanNguyen:
Can you review it, @vil02?

src/backtracking/graph_coloring.rs
```diff
@@ -0,0 +1,351 @@
...
+        ),
+        test_no_valid_coloring: (
+            vec![
+                vec![false, true, true],
+                vec![true, false, true],
+                vec![true, true, false],
+            ],
+            2,
+            Ok(None::<Vec<Vec<usize>>>)
+        ),
```
vil02, line 283:
```suggestion
        ),
        test_more_colors_than_nodes: (
            vec![
                vec![true, true],
                vec![true, true],
            ],
            3,
            Ok(Some(vec![
                vec![1, 2],
                vec![1, 3],
                vec![2, 1],
                vec![2, 3],
                vec![3, 1],
                vec![3, 2],
            ]))
        ),
        test_no_coloring_with_zero_colors: (
            vec![
                vec![true],
            ],
            0,
            Ok(None::<Vec<Vec<usize>>>)
        ),
```

src/backtracking/graph_coloring.rs
```diff
@@ -0,0 +1,351 @@
...
+    ///
+    /// * `vertex` - The current vertex to be colored.
+    /// * `num_colors` - The number of colors available for coloring the graph.
+    fn find_colorings(&mut self, vertex: usize, num_colors: usize) {
+        if vertex == self.num_vertices() {
+            self.solutions.push(self.vertex_colors.clone());
+            return;
+        }
+
+        for color in 1..=num_colors {
```
vil02, line 111: Is there a reason why you have chosen `1` to be the first color? It seams to be more natural (in languages like rust) to start with 0.

vil02 (changes requested):

src/backtracking/graph_coloring.rs
TruongNhanNguyen: It's a good idea.

src/backtracking/graph_coloring.rs
TruongNhanNguyen: There are no special reasons, I just use it on my own. After reviewing your suggestion, I see that it is more natural to start counting from 0, and using 0 as the first color can also make the implementation more consistent with typical array indexing.

TruongNhanNguyen:
Let's have a final look, @vil02.

vil02 (approved):

PR #738, opened 2024-06-06
Fix uninlined format args

vil02:
# Pull Request Template

## Description

This PR adds [`uninlined_format_args`](https://rust-lang.github.io/rust-clippy/master/#/uninlined_format_args) into the clippy setup.

## Checklist:

### DIFF
--- a/src/ciphers/morse_code.rs
+++ b/src/ciphers/morse_code.rs
@@ -172,12 +172,7 @@ mod tests {
     #[test]
     fn decrypt_valid_character_set_invalid_morsecode() {
         let expected = format!(
-            "{}{}{}{} {}",
-            _UNKNOWN_MORSE_CHARACTER,
-            _UNKNOWN_MORSE_CHARACTER,
-            _UNKNOWN_MORSE_CHARACTER,
-            _UNKNOWN_MORSE_CHARACTER,
-            _UNKNOWN_MORSE_CHARACTER,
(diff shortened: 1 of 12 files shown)


siriak (approved):

src/ciphers/morse_code.rs
```diff
@@ -172,12 +172,7 @@ mod tests {
     #[test]
     fn decrypt_valid_character_set_invalid_morsecode() {
         let expected = format!(
-            "{}{}{}{} {}",
-            _UNKNOWN_MORSE_CHARACTER,
-            _UNKNOWN_MORSE_CHARACTER,
-            _UNKNOWN_MORSE_CHARACTER,
-            _UNKNOWN_MORSE_CHARACTER,
-            _UNKNOWN_MORSE_CHARACTER,
+            "{_UNKNOWN_MORSE_CHARACTER}{_UNKNOWN_MORSE_CHARACTER}{_UNKNOWN_MORSE_CHARACTER}{_UNKNOWN_MORSE_CHARACTER} {_UNKNOWN_MORSE_CHARACTER}",
```
siriak, line 175: The original version seems more legible to me, but it's up to you

siriak, line 175: The original version seems more legible to me, but it's up to you

vil02: Both are _ugly_. If the `_UNKNOWN_MORSE_CHARACTER` would be shorted, the 2nd version would look nicer. I do not see a better solution at the moment. I will merge it for not. Thanks for the review.

PR #741, opened 2024-06-10
Implemented Minkowski Difference

jerr-it:
## Description

The Minkowski distance or Minkowski metric is a metric in a normed vector space which can be considered as a generalization of both the Euclidean distance and the Manhattan distance.
It is named after the Polish mathematician Hermann Minkowski.
https://en.wikipedia.org/wiki/Minkowski_distance

This metric is useful in the field of machine learning to judge the similarity between data points.
It's also employed for collision detection in game engines.

## Type of change

## Checklist:

- [*] I ran `cargo clippy --all -- -D warnings` just before my last commit and fixed any issue that was found.

* The shown issues are unrelated to my changes and already existed beforehand


src/math/minkowski_difference.rs
```diff
@@ -0,0 +1,86 @@
...
+    pub x: Vec<f64>,
+}
+
+#[derive(Debug, PartialEq)]
+pub enum ParameterError {
+    DifferentDimensions,
+    InvalidOrder,
+}
+
+pub fn minkowski_difference(a: &Point, b: &Point, order: u32) -> Result<f64, ParameterError> {
```
vil02, line 20:
```suggestion
pub fn minkowski_distance(a: &Point, b: &Point, order: f64) -> Result<f64, ParameterError> {
```
There are two things:
- the name: why not `minkowski_distance`?
- why to limit `order` to `u32`?

src/math/minkowski_difference.rs
```diff
@@ -0,0 +1,86 @@
...
+    if order < 1 {
+        return Err(ParameterError::DifferentDimensions);
```
vil02, line 22:
```suggestion
        return Err(ParameterError::InvalidOrder);
```

src/math/minkowski_difference.rs
```diff
@@ -0,0 +1,86 @@
...
+        let a = Point { x: vec![1.0, 1.0] };
+        let b = Point {
+            x: vec![2.0, 2.0, 2.0],
+        };
+
+        assert_eq!(
+            minkowski_difference(&a, &b, 0),
+            Err(ParameterError::DifferentDimensions)
+        );
+    }
```
vil02, line 85:
```suggestion
    macro_rules! test_minkowski_difference {
        ($($name:ident: $test_case:expr,)*) => {
            $(
                #[test]
                fn $name() {
                    let (a, b, order, expected) = $test_case;
                    assert_eq!(minkowski_difference(&a, &b, order), expected);
                    assert_eq!(minkowski_difference(&b, &a, order), expected);
                }
            )*
        }
    }

    test_minkowski_difference! {
        pow_1: (Point { x: vec![1.0, 1.0] }, Point { x: vec![2.0, 2.0] }, 1, Ok(2.0)),
        different_dimensions: (Point { x: vec![1.0, 1.0] }, Point { x: vec![2.0, 2.0, 2.0] }, 1, Err(ParameterError::DifferentDimensions)),
        order_less_than_1: (Point { x: vec![1.0, 1.0] }, Point { x: vec![2.0, 2.0] }, 0, Err(ParameterError::InvalidOrder)),
    }
```

src/math/minkowski_difference.rs
```diff
@@ -0,0 +1,86 @@
+/// The Minkowski distance or Minkowski metric is a metric in a normed vector space which
+/// can be considered as a generalization of both the Euclidean distance and the Manhattan distance.
+/// It is named after the Polish mathematician Hermann Minkowski.
```
vil02, line 3: Polish? And maybe you will tell me that Hugo Steinhaus was German?

src/math/minkowski_difference.rs
```diff
@@ -0,0 +1,86 @@
...
+/// https://en.wikipedia.org/wiki/Minkowski_distance
+///
+/// This metric is useful in the field of machine learning to judge the similarity between data points.
+/// It's also employed for collision detection in game engines.
+use num_traits::Pow;
+
+pub struct Point {
+    pub x: Vec<f64>,
+}
```
vil02, line 12: I am not sure if this is really needed. I would be fine with using _naked_ `Vec<f64>`.

vil02 (changes requested):

PR #743, opened 2024-06-11
style: make clippy more strict

vil02:
# Pull Request Template

## Description

This PR makes clippy _more strict_. By adding `pedantic`, `restriction`, `nursery` and `cargo` lint groups to its config. Clearly some of the warnings need to be suppressed (the suppression list is _minimal_, i.e. removing any of if, without making any change in the code, will cause clippy to fail).

I think that most of the suppressed warnings, especially from the `pedantic` group, should be fixed in future.

Continuation of:
- #738
- #739

In the same spirit as:
- TheAlgorithms/Java#5110
- TheAlgorithms/Java#5122
- TheAlgorithms/Java#5126
- TheAlgorithms/Java#5144
- TheAlgorithms/Java#5155
- TheAlgorithms/Java#5165

## Checklist:


siriak (approved):
Looks good, thanks!

PR #745, opened 2024-06-12
Improve Ford Fulkerson Implementation

TruongNhanNguyen:
## Changes

- Added Rust doc comments to provide better documentation for the functions and the module
- Renamed variables for better readability
- Use a macro to define the tests and add some edge tests
- Add custom error types

## Checklist:


TruongNhanNguyen:
Let's take a look @vil02

TruongNhanNguyen:
Can you review this PR, @vil02?

src/graph/ford_fulkerson.rs
```diff
@@ -1,141 +1,295 @@
...
-const V: usize = 6; // Number of vertices in graph
+/// Enum representing the possible errors that can occur when running the Ford-Fulkerson algorithm.
+#[derive(Debug, PartialEq)]
+pub enum FordFulkersonError {
+    /// Error indicating that the graph is empty or has no edges.
+    EmptyGraph,
+    /// Error indicating that the source vertex is out of bounds.
+    InvalidSource,
+    /// Error indicating that the sink vertex is out of bounds.
+    InvalidSink,
```
vil02, line 16:
```suggestion
    SourceOutOfBounds,
    SinkOutOfBounds,
```

src/graph/ford_fulkerson.rs
```diff
@@ -1,141 +1,295 @@
...
-       iii. Increase the flow along the augmenting path by the determined amount.
-    3.Return the maximum flow.
-
-*/
 use std::collections::VecDeque;
 
-const V: usize = 6; // Number of vertices in graph
+/// Enum representing the possible errors that can occur when running the Ford-Fulkerson algorithm.
+#[derive(Debug, PartialEq)]
+pub enum FordFulkersonError {
```
vil02, line 10: What if the `graph` is not a square matrix?

vil02 (changes requested):

src/graph/ford_fulkerson.rs
TruongNhanNguyen: Just need to add an error variant to handle this case.

src/graph/ford_fulkerson.rs
```diff
@@ -1,141 +1,311 @@
...
+    /// Error indicating that the graph is empty or has no edges.
+    EmptyGraph,
+    /// Indicates that the graph is not a square matrix.
+    ImproperGraph,
+    /// Error indicating that the source vertex is out of bounds.
```
vil02, line 15:
```suggestion
```
Error name is descriptive enough.

src/graph/ford_fulkerson.rs
```diff
@@ -1,141 +1,311 @@
...
+    SourceOutOfBounds,
+    /// Error indicating that the sink vertex is out of bounds.
```
vil02, line 17:
```suggestion
```
Same here.

src/graph/ford_fulkerson.rs
```diff
@@ -1,141 +1,311 @@
...
+///
+/// # Returns
+///
+/// Returns the maximum flow and the residual graph
+pub fn ford_fulkerson(
+    graph: &[Vec<isize>],
+    source: usize,
+    sink: usize,
+) -> Result<isize, FordFulkersonError> {
+    if graph.is_empty() || graph[0].is_empty() {
```
vil02, line 75: I am not sure about this check, i.e. about the `graph[0].is_empty()`. If `graph[0]` exists, then `graph.len() > 0`, so if we only deal with _proper_ graphs (i.e. having square _flow_ matrix), then `graph[0].len()` must be `graph.len()`.

If you remove this part of the condition, please update the comment explaining the `FordFulkersonError::EmptyGraph`.

src/graph/ford_fulkerson.rs
```diff
@@ -1,141 +1,311 @@
...
+            ],
+            0,
+            6,
+            Err(FordFulkersonError::SinkOutOfBounds),
+        ),
+        test_improper_graph: (
+            vec![
+                vec![0, 8],
+                vec![0, 0, 9],
+                vec![0, 0, 0, 0, 7],
```
vil02, line 181: Why not to use a _smaller_ test case?

```suggestion
                vec![0, 8],
                vec![0],
```

src/graph/ford_fulkerson.rs
```diff
@@ -1,141 +1,311 @@
...
-        ];
-        assert_eq!(ford_fulkerson(&graph, 0, 5), 6);
+    macro_rules! test_max_flow {
+        ($($name:ident: $tc:expr,)* ) => {
+            $(
+                #[test]
+                fn $name() {
+                    let (graph, source, sink, expected_result) = $tc;
+                    let result = ford_fulkerson(&graph, source, sink);
+                    assert_eq!(result, expected_result);
```
vil02, line 130:
```suggestion
                    assert_eq!(ford_fulkerson(&graph, source, sink), expected_result);
```

src/graph/ford_fulkerson.rs
```diff
@@ -1,141 +1,311 @@
...
+        return Err(FordFulkersonError::ImproperGraph);
+    }
+
+    if source >= graph.len() {
+        return Err(FordFulkersonError::SourceOutOfBounds);
+    }
+
+    if sink >= graph.len() {
+        return Err(FordFulkersonError::SinkOutOfBounds);
+    }
```
vil02, line 89:
```suggestion
    check_input(graph, source, sink)?;
```
with
```rust
fn check_input(graph: &[Vec<isize>], source: usize, sink: usize) -> Result<(), FordFulkersonError> {
    if graph.is_empty() || graph[0].is_empty() { // please take care of the "|| graph[0].is_empty()" part
        return Err(FordFulkersonError::EmptyGraph);
    }

    if graph.iter().any(|row| row.len() != graph.len()) {
        return Err(FordFulkersonError::ImproperGraph);
    }

    if source >= graph.len() {
        return Err(FordFulkersonError::SourceOutOfBounds);
    }

    if sink >= graph.len() {
        return Err(FordFulkersonError::SinkOutOfBounds);
    }
    Ok(())
}
```

src/graph/ford_fulkerson.rs
```diff
@@ -1,141 +1,311 @@
...
 
-pub fn ford_fulkerson(graph: &[Vec<i32>], s: usize, t: usize) -> i32 {
-    let mut r_graph = graph.to_owned();
-    let mut parent = vec![-1; V];
+/// Applies the Ford-Fulkerson algorithm to find the maximum flow in a flow network
+/// represented by a weighted directed graph.
+///
+/// # Arguments
+///
+/// * `graph` - A mutable reference to the flow network represented as an adjacency matrix.
```
vil02, line 63:
```suggestion
/// * `graph` - A reference to the flow network represented as an adjacency matrix.
```

vil02 (changes requested):

src/graph/ford_fulkerson.rs
TruongNhanNguyen: If the graph is a square matrix, which it should be for a proper flow network, then `graph[0].len()` will always be equal to `graph.len()` if `graph` is non-empty. As a result, the check for `graph[0].is_empty()` is unnecessary. We can simplify the condition and update the error handling accordingly

src/graph/ford_fulkerson.rs
TruongNhanNguyen: Just use it now.

src/graph/ford_fulkerson.rs
TruongNhanNguyen: I will try this one

src/graph/ford_fulkerson.rs
TruongNhanNguyen: Will update the docs

src/graph/ford_fulkerson.rs
TruongNhanNguyen: Yes, sometimes I overuse comments

src/graph/ford_fulkerson.rs
```diff
@@ -1,141 +1,324 @@
...
+            0,
+            0,
+            Ok(0),
+        ),
+        test_negative_capacity: (
+            vec![
+                vec![0, -10, 0],
+                vec![0, 0, -5],
+                vec![0, 0, 0],
+            ],
```
vil02, line 299: This is something what bothers me: do we really want to have inputs with negative entries?

TruongNhanNguyen: Because the graph is `Vec<Vec<isize>>`, I think it's not a redundant test. What do you think about it?

TruongNhanNguyen: If you feel this test is not necessary, feel free to remove it. I added this test because the implementation should handle the cases where the graph has negative flows.

vil02: First of all, I am not sure if negative numbers make any sense here. Secondly, please have a look at this line:
https://github.com/TheAlgorithms/Rust/blob/2576c211e200deb8f1998a71a1f32dd71608d9cb/src/graph/ford_fulkerson.rs#L40

What should be the output for
```rust
vec![
    vec![0, -1, 1],
    vec![-1, 0, 100],
    vec![1, 100, 0],
],
```
99 or 1?

TruongNhanNguyen: Yes, I overlook it. The negative flows do not make any sense in this implementation.

TruongNhanNguyen: Because the flow always is non-negative values, we can make the following changes:
- The graph representation should be `Vec<Vec<usize>`
- The parent array in the Ford-Fulkerson algorithm is used to store the augmenting path found by the BFS and should be an array of `usize` values (eliminate type casting from `isize` to `usize`)
- The tests for negative cases should be removed (it does not make any sense in the implementation)

TruongNhanNguyen: Let's see the change @vil02?

vil02 (approved):

PR #766, opened 2024-07-13
Improve binary search tree implementation

DaevMithran:
## Changes

- Implement binary search tree without recursion

## Checklist:

### DIFF
--- a/src/data_structures/binary_search_tree.rs
+++ b/src/data_structures/binary_search_tree.rs
@@ -3,6 +3,7 @@ use std::ops::Deref;
 
 /// This struct implements as Binary Search Tree (BST), which is a
 /// simple data structure for storing sorted data
+#[derive(Debug)]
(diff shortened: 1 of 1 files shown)


src/data_structures/binary_search_tree.rs
```diff
@@ -3,6 +3,7 @@ use std::ops::Deref;
 
 /// This struct implements as Binary Search Tree (BST), which is a
 /// simple data structure for storing sorted data
+#[derive(Debug)]
```
vil02, line 6: Is this needed?

```suggestion
```

src/data_structures/binary_search_tree.rs
vil02, line 1: Using this opportunity, could you please add the _missing tests_? There are few lines, which are not cover. It would be great if you could provide test cases _covering_ these scenarios.

vil02 (changes requested):

PR #767, opened 2024-07-15
perf: Optimize BTree search using binary search

ie-Yoshisaur:
## Description
This pull request optimizes the search function in the BTree implementation by replacing the linear search with a binary search algorithm. This change significantly improves the search performance, especially for large trees, without altering the existing functionality.

The optimization is achieved by modifying the `search` method in the `BTree` struct to use the `binary_search` method on the `keys` vector instead of a while loop. This change reduces the time complexity of searching within a node from O(n) to O(log n), where n is the number of keys in the node.

## Type of change

## Implementation Details
The main changes are in the `search` method of the `BTree` struct:
1. Replaced the while loop with `binary_search` method on the `keys` vector.

## Benchmark Results
Environment: MacOS 14.5, Apple M1 Pro, 32 GB RAM
Dataset: 1,000,000 random integers for insertion, 1,000,000 searches for the key 500,000

Before optimization:
- Insertion: 3.002587333s
- Search: 2.334683584s

After optimization:
- Insertion: 2.998482583s
- Search: 288.659458ms

Note: Insertion time remains largely unchanged, as expected.

## Checklist:
- [x] I ran the following commands using the latest version of **rust nightly**.

All existing tests pass with the new implementation. No new tests were added as the existing ones adequately cover the modified functionality.


vil02 (approved):
Thanks for spotting that. I added few tests: b772765011c3a407f9272362a2b19b756014d950.

By the way: would you like to spend some more time with this class? Code like `BTree::new(0)` or `BTree::new(1)` should not work in my opinion. Feel free to open a new PR fixing that.

ie-Yoshisaur:
> By the way: would you like to spend some more time with this class?

Of course! I’ll work on fixing that issue and submit a PR soon.

PR #771, opened 2024-07-23
Improve Longest Common Subsequence Implementation

TruongNhanNguyen:
## Changes made

- Decompose the `longest_common_subsequence` to use helper functions
- Rewrite tests using macro
- Naming variables and functions descriptively
- Add docstrings

## Checklist:

### DIFF
--- a/src/dynamic_programming/longest_common_subsequence.rs
+++ b/src/dynamic_programming/longest_common_subsequence.rs
@@ -9,9 +9,9 @@
 ///
 /// The longest common subsequence (LCS) of two strings is the longest sequence that can
 /// be derived from both strings by deleting some elements without changing the order of
-/// the remaining elements. When there are multiple possible LCSs of the same length, the specific LCS returned
-/// depends on the order of the input sequences and how ties are resolved in the dynamic
-/// programming table.
+/// the remaining elements. When there are multiple possible LCSs of the same length,
+/// the specific LCS returned depends on the order of the input sequences and how ties
+/// are resolved in the dynamic programming table.
 ///
 /// ## Note
 /// The function may return different LCSs for the same pair of strings depending on the
@@ -31,7 +31,6 @@ pub fn longest_common_subsequence(first_seq: &str, second_seq: &str) -> String {
     lcs_chars.into_iter().collect()
 }
 
-/// Initializes the LCS length table using dynamic programming.
 fn initialize_lcs_lengths(first_seq_chars: &[char], second_seq_chars: &[char]) -> Vec<Vec<usize>> {
     let first_seq_len = first_seq_chars.len();
     let second_seq_len = second_seq_chars.len();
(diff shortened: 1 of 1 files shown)


TruongNhanNguyen:
Let's have a look, @vil02

src/dynamic_programming/longest_common_subsequence.rs
vil02, line 1: Could you please document _which_ LCS is returned? For example:
`longest_common_subsequence("hello, world!", "world, hello!")` returns `"hello!"`
but
`longest_common_subsequence("world, hello!", "hello, world!")` returns `"world!"`.

vil02 (changes requested):

src/dynamic_programming/longest_common_subsequence.rs
TruongNhanNguyen: I have explained the LCS in the latest commit. You can see it now.

TruongNhanNguyen: Can you have a look, @vil02.

vil02: Could you try to explain _why_ `longest_common_subsequence("hello, world!", "world, hello!")` returns `"hello!"`
and _why_
`longest_common_subsequence("world, hello!", "hello, world!")` returns `"world!"`?

TruongNhanNguyen: I understand your idea and try to explain the differences in the best way.
```rust
//! This module implements the Longest Common Subsequence (LCS) algorithm.
//! The LCS problem is finding the longest subsequence common to two sequences.
//! It differs from the problem of finding common substrings: unlike substrings, subsequences
//! are not required to occupy consecutive positions within the original sequences.
//! This implementation handles Unicode strings efficiently and correctly, ensuring
//! that multi-byte characters are managed properly.

/// Computes the longest common subsequence of two input strings.
///
/// The longest common subsequence (LCS) of two strings is the longest sequence that can
/// be derived from both strings by deleting some elements without changing the order of
/// the remaining elements.
///
/// ## Note
/// The function may return different LCSs for the same pair of strings depending on the
/// order of the inputs and the nature of the sequences. This is due to the way the dynamic
/// programming algorithm resolves ties when multiple common subsequences of the same length
/// exist. The order of the input strings can influence the specific path taken through the
/// DP table, resulting in different valid LCS outputs.
///
///  For example:
/// `longest_common_subsequence("hello, world!", "world, hello!")` returns `"hello!"`
/// but
/// `longest_common_subsequence("world, hello!", "hello, world!")` returns `"world!"`
///
/// This difference arises because the dynamic programming table is filled differently based
/// on the input order, leading to different tie-breaking decisions and thus different LCS results.
pub fn longest_common_subsequence(first_seq: &str, second_seq: &str) -> String {
}
```

To summarize, the difference in output is due to the order of input strings and the specific way the DP algorithm handles ties when constructing the LCS. The same sequence pairs can lead to different LCS results depending on how the paths through the DP table are chosen based on input order and the algorithm's tie-breaking behavior.

TruongNhanNguyen: Do you have any suggestion, @vil02?

vil02 (approved):

PR #772, opened 2024-07-25
Improve infix to postfix notation algorithm

bgdnrvsky:
## Description

- Moved the algorithm to a more appropriate place
- Introduced the error handling for the algorithm
- Added erroneous unit tests

## Type of change

## Checklist:


src/conversions/infix_to_postfix.rs
```diff
@@ -1,10 +1,16 @@
...
+}
+
+/// Function to convert [infix expression](https://en.wikipedia.org/wiki/Infix_notation) to [postfix expression](https://en.wikipedia.org/wiki/Reverse_Polish_notation)
+pub fn infix_to_postfix(infix: &str) -> Result<String, InfixToPostfixError> {
     let mut postfix = String::new();
     let mut stack: Vec<char> = Vec::new();
 
     // Define the precedence of operators
-    let precedence = |op: char| -> i32 {
+    let precedence = |op: char| -> u8 {
```
vil02, line 13: What are the reasons to change from `i32` to `u8`? Just saving these 3 bytes? From the performance point of view `i32` could be faster. I do not have an opinion here, I am just asking.

src/conversions/infix_to_postfix.rs
vil02, line 1: I am not sure if moving this file to `conversions` is a good idea. I guess it was in `data_structures`, because it used stack. It could well be in `math`. In any case, i would suggest it to leave it where it was.

vil02 (changes requested):

src/conversions/infix_to_postfix.rs
bgdnrvsky: The reason why I changed it, in the first place was the fact that, if someone wants to use this algorithm, he will probably need to adapt it, so while the change doesn't _really_ affect the algorithm, it makes the user understand that there is no negative precedence (AFAIK)

src/conversions/infix_to_postfix.rs
```diff
@@ -38,19 +44,19 @@ pub fn infix_to_postfix(infix: &str) -> String {
...
         }
+
         postfix.push(top);
     }
 
-    postfix
+    Ok(postfix)
 }
 
 #[cfg(test)]
```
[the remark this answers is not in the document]
bgdnrvsky: Sure, why not

src/conversions/infix_to_postfix.rs
bgdnrvsky: I don't see how the fact that it uses a stack can be a reason for it to be in `data_structures`, following this logic, any algorithm would be here, no ? And if you look at the `data_structures` section, you can clearly see the members that should certainly be in there, like `HashTable`, `Heap`, `Graph`, etc . I think that the best solution would be to put it, as you mentioned it, in the `math`.

vil02: I agree with `math`.

vil02 (approved):

vil02:
There was an update of clippy and some warnings become active. Could you just _suppress_ them in [`Cargo.toml`](https://github.com/TheAlgorithms/Rust/blob/b8fa915a59a44ffc09fd46c3eaf0ca1a3902cecd/Cargo.toml)?

vil02 (approved):

PR #778, opened 2024-08-11
Refactor palindrom checking algorithm

bgdnrvsky:
## Description

- Use simpler approach to check if the string is a palindrome
- Add unit test for empty palindrome

## Checklist:

### DIFF
--- a/src/string/palindrome.rs
+++ b/src/string/palindrome.rs
@@ -1,11 +1,5 @@
 pub fn is_palindrome(s: &str) -> bool {
-    let mut chars = s.chars();
-    while let (Some(c1), Some(c2)) = (chars.next(), chars.next_back()) {
-        if c1 != c2 {
-            return false;
-        }
-    }
-    true
+    s.chars().eq(s.chars().rev())
 }
 
 #[cfg(test)]
--- a/src/string/palindrome.rs
+++ b/src/string/palindrome.rs
@@ -8,6 +8,7 @@ mod tests {
 
     #[test]
     fn palindromes() {
+        assert!(is_palindrome(""));
         assert!(is_palindrome("abcba"));
         assert!(is_palindrome("abba"));
         assert!(is_palindrome("a"));


src/string/burrows_wheeler_transform.rs
```diff
@@ -42,76 +42,55 @@ pub fn inv_burrows_wheeler_transform<T: AsRef<str>>(input: (T, usize)) -> String
...
+            burrows_wheeler_transform($string)
+        };
+    }
+
+    macro_rules! decipher {
+        ($string:expr) => {
+            inv_burrows_wheeler_transform($string)
+        };
+    }
+
```
vil02, line 56: What is the advantage of using these macros?

src/string/burrows_wheeler_transform.rs
vil02, line 1: The title of this PR is _Refactor palindrom checking algorithm_. In order to move faster, I would suggest to revert the changes of this file and create another PR with them.

src/string/palindrome.rs
```diff
@@ -1,11 +1,5 @@
 pub fn is_palindrome(s: &str) -> bool {
-    let mut chars = s.chars();
-    while let (Some(c1), Some(c2)) = (chars.next(), chars.next_back()) {
-        if c1 != c2 {
-            return false;
-        }
-    }
-    true
+    s.chars().eq(s.chars().rev())
```
vil02, line 2: This is a nice one liner but quite inefficient (both in terms of space and time). Would you like to rewrite it avowing the copy and avoiding the redundant checks in case of a palindrome?

vil02 (changes requested):

src/string/burrows_wheeler_transform.rs
bgdnrvsky: Yeah, sorry, accidentally pushed it to the `master` branch

src/string/palindrome.rs
bgdnrvsky: Agree with you, at this point it is better to close the PR

PR #782, opened 2024-08-24
Improve All Combinations of Size k Implementation

sozelfist:
## Checklist:

### DIFF
--- a/src/backtracking/all_combination_of_size_k.rs
+++ b/src/backtracking/all_combination_of_size_k.rs
@@ -1,33 +1,47 @@
-/*
-    In this problem, we want to determine all possible combinations of k
-    numbers out of 1 ... n. We use backtracking to solve this problem.
-    Time complexity: O(C(n,k)) which is O(n choose k) = O((n!/(k! * (n - k)!)))
+//! This module provides a function to generate all possible combinations
+//! of `k` numbers out of `1...n` using a backtracking algorithm.
 
-    generate_all_combinations(n=4, k=2) => [[1, 2], [1, 3], [1, 4], [2, 3], [2, 4], [3, 4]]
-*/
-pub fn generate_all_combinations(n: i32, k: i32) -> Vec<Vec<i32>> {
-    let mut result = vec![];
-    create_all_state(1, n, k, &mut vec![], &mut result);
-
-    result
+/// Generates all possible combinations of `k` numbers out of `1...n`.
+///
+/// # Arguments
+///
+/// * `n` - The upper limit of the range (1 to `n`).
+/// * `k` - The number of elements in each combination.
+///
+/// # Returns
+///
+/// A vector containing all possible combinations of `k` numbers out of `1...n`.
+pub fn generate_all_combinations(n: isize, k: isize) -> Vec<Vec<isize>> {
+    let mut combinations = vec![];
+    backtrack(1, n, k, &mut vec![], &mut combinations);
+    combinations
 }
 
-fn create_all_state(
-    increment: i32,
-    total_number: i32,
-    level: i32,
-    current_list: &mut Vec<i32>,
-    total_list: &mut Vec<Vec<i32>>,
+/// Helper function to generate combinations recursively.
+///
+/// # Arguments
+///
+/// * `start` - The current number to start the combination with.
+/// * `n` - The upper limit of the range (1 to `n`).
+/// * `k` - The number of elements left to complete the combination.
+/// * `current` - A mutable reference to the current combination being constructed.
+/// * `combinations` - A mutable reference to the vector holding all combinations.
+fn backtrack(
+    start: isize,
+    n: isize,
+    k: isize,
+    current: &mut Vec<isize>,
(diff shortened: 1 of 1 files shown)


sozelfist:
Can you have a look, @vil02?

src/backtracking/all_combination_of_size_k.rs
```diff
@@ -1,62 +1,121 @@
...
+    }
+
+    if k > n {
+        return Err(CombinationError::KGreaterThanN);
+    }
+
+    let mut combinations = vec![];
+    backtrack(1, n, k, &mut vec![], &mut combinations);
+    Ok(combinations)
 }
```
vil02, line 34: What do you think abut optimizing the code a bit by avoiding pushes to `current`? Namely
```rust
pub fn generate_all_combinations(n: usize, k: usize) -> Result<Vec<Vec<usize>>, CombinationError> {
    if n == 0 && k > 0 {
        return Err(CombinationError::InvalidZeroRange);
    }

    if k > n {
        return Err(CombinationError::KGreaterThanN);
    }

    let mut combinations = vec![];
    let mut current = vec![0; k];
    backtrack(1, n, k, 0, &mut current, &mut combinations);
    Ok(combinations)
}


fn backtrack(
    start: usize,
    n: usize,
    k: usize,
    index: usize,
    current: &mut Vec<usize>,
    combinations: &mut Vec<Vec<usize>>,
) {
    if index == k {
        combinations.push(current.clone());
        return;
    }

    for num in start..=(n - k + index + 1) {
        current[index] = num;
        backtrack(num + 1, n, k, index + 1, current, combinations);
    }
}
```

src/backtracking/all_combination_of_size_k.rs
```diff
@@ -1,62 +1,121 @@
...
-
-    generate_all_combinations(n=4, k=2) => [[1, 2], [1, 3], [1, 4], [2, 3], [2, 4], [3, 4]]
-*/
-pub fn generate_all_combinations(n: i32, k: i32) -> Vec<Vec<i32>> {
-    let mut result = vec![];
-    create_all_state(1, n, k, &mut vec![], &mut result);
-
-    result
+//! This module provides a function to generate all possible combinations
+//! of `k` numbers out of `1...n` using a backtracking algorithm.
```
vil02, line 2: I think it would be more natural to have:
```suggestion
//! of `k` numbers out of `0...n-1` using a backtracking algorithm.
```

What do you think?

vil02 (changes requested):

src/backtracking/all_combination_of_size_k.rs
sozelfist: Yes, It would be more natural when using `0...n-1` instead of `1...n`. Especially since it's consistent with how ranges are commonly used in programming languages.

src/backtracking/all_combination_of_size_k.rs
sozelfist: This optimization is a good idea. By pre-allocating a `current` vector and using an index to track the position instead of repeatedly pushing and popping elements, the code becomes more efficient. So we can avoid unnecessary heap allocations during each recursive call. It helps us reduce memory usage and slightly improve the performance of the algorithm.

Great point here.

src/backtracking/all_combination_of_size_k.rs
```diff
@@ -1,62 +1,121 @@
...
+    if n == 0 && k > 0 {
+        return Err(CombinationError::InvalidZeroRange);
+    }
+
+    if k > n {
+        return Err(CombinationError::KGreaterThanN);
+    }
+
+    let mut combinations = vec![];
+    backtrack(1, n, k, &mut vec![], &mut combinations);
```
sozelfist, line 32:
```suggestion
    let mut current = vec![0; k];
    backtrack(0, n, k, 0, &mut current, &mut combinations); // Start from 0 for zero-indexing
```
The recursion starts from `start = 0`, so it aligns with zero-indexing.

src/backtracking/all_combination_of_size_k.rs
```diff
@@ -1,62 +1,121 @@
...
-    for i in increment..(total_number - level + 2) {
-        current_list.push(i);
-        create_all_state(i + 1, total_number, level - 1, current_list, total_list);
-        current_list.pop();
+    for num in start..=(n - k + 1) {
+        current.push(num);
+        backtrack(num + 1, n, k - 1, current, combinations);
+        current.pop();
     }
 }
```
sozelfist, line 62:
```rs
fn backtrack(
    start: usize,
    n: usize,
    k: usize,
    index: usize,
    current: &mut Vec<usize>,
    combinations: &mut Vec<Vec<usize>>,
) {
    if index == k {
        combinations.push(current.clone());
        return;
    }

    for num in start..=(n - k + index) {  // Adjusted the range to reflect zero-based indexing
        current[index] = num;
        backtrack(num + 1, n, k, index + 1, current, combinations);
    }
}
```

vil02 (approved):

PR #783, opened 2024-08-26
Added is_valid_email_address.rs

lyqio:
# Changes

- add is_valid_email_address.rs

## Type of change

## Checklist:

### DIFF
--- a/src/string/is_valid_email_address.rs
+++ b/src/string/is_valid_email_address.rs
@@ -1,0 +1,303 @@
+const MAX_LOCAL_PART_LENGTH: u32 = 64;
+const MAX_DOMAIN_LENGTH: u32 = 255;
+const CHARACTERS: &str = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz1234567890";
+const SYMBOLS: &str = "!#$%&'*+-/=?^_`{|}~";
+const HEX_CHARACTERS: &str = "0123456789abcdef";
+
+fn clear_quotes(input: &str) -> String {
+    let mut quote_count = input.starts_with('"') as u32;
+    let mut new_local_part: String = String::from(input.chars().next().unwrap());
+    for i in 1..input.len() {
+        if input.chars().nth(i).unwrap() == '"' && input.chars().nth(i - 1).unwrap() != '\\' {
+            quote_count += 1;
+
+            if !new_local_part.starts_with('"') && quote_count != 1 {
+                new_local_part.push('"');
+            }
+        }
+
+        if quote_count % 2 == 0 {
+            new_local_part.push(input.chars().nth(i).unwrap());
+        }
+    }
+
+    new_local_part
+}
+
+fn has_bad_dots(part: &str) -> bool {
+    part.starts_with('.') || part.ends_with('.') || part.contains("..")
+}
+
+fn contains_legal_local_characters(local_part: &str) -> bool {
+    for item in local_part.chars() {
+        if !CHARACTERS.contains(item) && !SYMBOLS.contains(item) && item != '.' && item != '"' {
+            return false;
+        }
+    }
+    true
+}
+
+fn contains_legal_domain_characters(domain: &str) -> bool {
+    for item in domain.chars() {
+        if !CHARACTERS.contains(item) && item != '-' && item != '.' {
+            return false;
+        }
+    }
+    true
+}
+
+fn is_valid_quotes(part: &str) -> bool {
+    for i in 1..part.len() {
+        if part.chars().nth(i - 1).unwrap() == '"' && part.chars().nth(i).unwrap() == '"' {
+            let proceeding_quote = part.chars().nth(i + 1);
+            if proceeding_quote.is_some() && proceeding_quote.unwrap() != '.' {
+                return false;
+            }
+        }
+    }
+    true
+}
+
+fn is_ipv4(ip: &str) -> bool {
+    let parts = ip
+        .split('.')
+        .map(|x| x.to_string())
+        .collect::<Vec<String>>();
+    if parts.len() != 4 {
+        return false;
+    }
+
+    for value in parts {
+        let value = value.parse::<u32>();
+        if value.is_err() || value.unwrap() > 255 {
+            return false;
+        }
+    }
+
+    true
+}
+
+fn is_ipv6(ip: &str) -> bool {
+    let parts = ip
+        .split(':')
+        .map(|x| x.to_string())
+        .collect::<Vec<String>>();
+
+    // we check for the length 9 here as the IPv6 counts as one of the parts
+    if parts.len() != 9 || parts.first().unwrap() != "IPv6" {
+        return false;
+    }
+
+    for value in &parts[1..parts.len()] {
+        if value.is_empty() || value.len() > 4 {
+            return false;
+        }
+
+        for chr in value.chars() {
+            if !HEX_CHARACTERS.contains(chr.to_ascii_lowercase()) {
+                return false;
+            }
+        }
+    }
+
+    true
+}
+
+pub fn is_valid_ip(domain: &str) -> bool {
+    is_ipv4(domain) || is_ipv6(domain)
+}
+
+fn is_valid_local_part(local_part: &str) -> bool {
+    if local_part.len() > MAX_LOCAL_PART_LENGTH as usize {
+        return false;
+    }
+
+    if !is_valid_quotes(local_part) {
+        return false;
+    }
+
+    if !contains_legal_local_characters(local_part) {
+        return false;
+    }
+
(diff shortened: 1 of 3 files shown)


vil02 (changes requested):
The definition of the _valid_ email address is slightly more involved, cf. https://en.wikipedia.org/wiki/Email_address#Examples
in particular, `"john..doe"@example.org` is perfectly fine.

This can be merged only if all of the _edge cases_ are covered.

lyqio (commented):
I rewrote some of the code and now passes all of the tests on https://en.wikipedia.org/wiki/Email_address#Examples

src/string/is_valid_email_address.rs
```diff
@@ -0,0 +1,139 @@
...
+    if domain.starts_with('.') || domain.ends_with('.') || domain.contains("..") {
+        return false;
+    }
+
+    // (8)  validate the dashes in the domain
+    if domain.starts_with('-') || domain.ends_with('-') {
+        return false;
+    }
+
+    true
```
vil02, line 91: What do you think about introducing functions like
- `is_local_part_valid`, which would then split into `is_quoted_local_part_valid` and `is_unquoted_local_part_valid`,
- `is_domain_valid`, which would then split accordingly.

src/string/is_valid_email_address.rs
```diff
@@ -0,0 +1,139 @@
...
+        slashes: ("name/surname@example.com", true),
+        no_tld: ("admin@example", true),
+        tld: ("example@s.example", true),
+        quotes_with_space: ("\" \"@example.org", true),
+        quoted_double_dot: ("\"john..doe\"@example.org", true),
+        host_route: ("mailhost!username@example.org", true),
+        quoted_non_letters: (r#""very.(),:;<>[]\".VERY.\"very@\\ \"very\".unusual"@strange.example.com"#, true),
+        percent_symbol: ("user%example.com@example.org", true),
+        local_end_symbol: ("user-@example.org", true),
+        ip_address: ("postmaster@[123.123.123.123]", true),
```
vil02, line 126:
```suggestion
        ip_address: ("postmaster@[123.123.123.123]", true),
        wrong_ip_address: ("postmaster@[1234.123.123.123]", false),
```

src/string/is_valid_email_address.rs
vil02, line 1: Please make sure that _tests cover all of the lines_.

vil02 (changes requested):

src/string/is_valid_email_address.rs
lyqio: I added checks for ip addresses and added some extra tests to ensure that the ip checks work as expected.

src/string/is_valid_email_address.rs
lyqio: Due to the way the code is written, checking the quoted part and unquoted part separately wouldn't make the code more readable, so instead I just refactored the code and split each piece of logic into smaller sections.

lyqio:
I handled all the things that you requested.

src/string/is_valid_email_address.rs
```diff
@@ -0,0 +1,213 @@
...
+        for section in sections {
+            let value = section.parse::<u32>();
+            if value.is_err() || value.unwrap() > 255 {
+                return false;
+            }
+        }
+    }
+
+    true
+}
```
vil02, line 90: Why not to split it into two functions: `is_ipv4` and `is_ipv6`? This will allow for testing them directly (currently lots of testes are missing e.g. `123.123.123.123.123.123.123.123` is not a valid ip address).

Furthermore, the logic here is quite convoluted. What do you think about checking (in `is_domain_valid`) if the _domain part_ has the form `[<some_str>]`, and if so, send the `<some_str>` to `is_ip` function. It could look like
```rust
fn is_valid_ip(ip_str: &str) -> bool {
    is_ipv4(ip_str) || is_ipv6(ip_str)
}
```

Such approach would me more natural: what is a valid ip address `255.255.255.255` or `[255.255.255.255]`?

vil02 (changes requested):

vil02:
Please make sure to update your branch (cf. #789) - otherwise the CI will fail.

lyqio:
I think I fixed all the issues and made the IP checking better.

src/string/is_valid_email_address.rs
```diff
@@ -0,0 +1,253 @@
...
+        if !domain.ends_with(']') {
+            return false;
+        }
+
+        let ip = &domain[1..domain.len() - 1];
+        return is_valid_ipv4(ip) || is_valid_ipv6(ip);
+    }
+
+    true
+}
```
vil02, line 121: This is quite convoluted and incorrect. Please follow the [earlier suggestions](https://github.com/TheAlgorithms/Rust/pull/783#discussion_r1759347825).

Note: `is_valid_email_address("postmaster@[")` returns `true`.

Also: `is_valid_ip("[]")` and `is_valid_ip("a")` would also return `true`.

vil02 (changes requested):

src/string/is_valid_email_address.rs
lyqio: I fixed the code and changed the is_valid_ip to make it fit your suggestion, It now passes the tests you provided.

src/string/is_valid_email_address.rs
```diff
@@ -0,0 +1,255 @@
...
+
+        for chr in value.chars() {
+            if !HEX_CHARACTERS.contains(chr.to_ascii_lowercase()) {
+                return false;
+            }
+        }
+    }
+
+    true
+}
```
vil02, line 104: What do you think about testing these functions directly?

src/string/is_valid_email_address.rs
```diff
@@ -0,0 +1,255 @@
...
+    {
+        return false;
+    }
+
+    if domain.starts_with('-') || domain.ends_with('-') {
+        return false;
+    }
+
+    true
+}
```
vil02, line 164: This could be simplified to
```rust
fn is_domain_valid(domain: &str) -> bool {
    if domain.starts_with('[') && domain.ends_with(']')
    {
        return is_valid_ip(&domain[1..domain.len() - 1])
    }

    if domain.len() > MAX_DOMAIN_LENGTH as usize {
        return false;
    }

    if !contains_legal_domain_characters(domain) {
        return false;
    }

    if has_bad_dots(domain) {
        return false;
    }

    if domain.starts_with('[') || domain.ends_with(']') {
        return false
    }

    if domain.starts_with('-') || domain.ends_with('-') {
        return false;
    }

    true
}
```
But it can be simplified even further: for example when you know that you are not dealing with the ip address, you know that a valid domain will not contain any bracket.

vil02 (changes requested):
Please consider using more consistent naming: sometime the `valid` part is in the middle and sometimes at the end. Perhaps it is not needed at all?

src/string/is_valid_email_address.rs
lyqio: I renamed `is_valid_ipv4` and `is_valid_ipv6` to `is_ipv4` and `is_ipv6` respectively, but for the rest of the functions which contained the word 'valid', I moved the valid to the middle of the function name to make the code more consistent while staying readable.

In addition I added tests for `is_ipv4` and `is_ipv6`, note that some ips were taken from the `is_valid_email_address` tests but I still kept those tests in as I thought they may still be good tests to have, they can be removed if you want. Also, due to adding more tests, I made two new macros, but the way I did it may not be the most efficient way of doing so, but it shouldn't matter too much as it is just for the tests.

src/string/is_valid_email_address.rs
lyqio: I simplified the code as previously mentioned

src/string/is_valid_email_address.rs
```diff
@@ -0,0 +1,303 @@
...
+    if !contains_legal_local_characters(local_part) {
+        return false;
+    }
+
+    if has_bad_dots(local_part) {
+        return false;
+    }
+
+    true
+}
```
vil02, line 128: What do you think about having a similar approach as in `is_valid_domain`? I.e. if the `local_part` starts and ends `"`. then a function `is_valid_local_quoted_part` and `is_valid_local_unquoted_part`?

lyqio: I don't think this would work as not all email addresses are made up of only a single string, for example `a."b".c` is valid although it does not begin and end with quotes. In addition due to the fact that the contents of all strings are removed at the start of the local part checking, I believe that it would make the code less readable.

vil02: `a."b".c` is _unquoted_. It contains a `"` character, therefore it is not valid (cf. [wiki](https://en.wikipedia.org/wiki/Email_address#Local-part)).

PR #788, opened 2024-09-12
Improve bit manipulation algorithm implementations

sozelfist:
## Changes made:
- Add docstring to file and function
- Rewrite tests using macro

## Checklist:

### DIFF
--- a/src/bit_manipulation/counting_bits.rs
+++ b/src/bit_manipulation/counting_bits.rs
@@ -1,11 +1,19 @@
-/*
-The counting bits algorithm, also known as the "population count" or "Hamming weight,"
-calculates the number of set bits (1s) in the binary representation of an unsigned integer.
-It uses a technique known as Brian Kernighan's algorithm, which efficiently clears the least
-significant set bit in each iteration.
-*/
+//! This module implements a function to count the number of set bits (1s)
+//! in the binary representation of an unsigned integer.
+//! It uses Brian Kernighan's algorithm, which efficiently clears the least significant
+//! set bit in each iteration until all bits are cleared.
+//! The algorithm runs in O(k), where k is the number of set bits.
 
-pub fn count_set_bits(mut n: u32) -> u32 {
+/// Counts the number of set bits in an unsigned integer.
+///
+/// # Arguments
+///
+/// * `n` - An unsigned 32-bit integer whose set bits will be counted.
+///
+/// # Returns
+///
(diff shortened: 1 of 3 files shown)


vil02:
> > Note: the CI Clippy step failed because of the new stable rust version

This should be handled by #789.

src/bit_manipulation/counting_bits.rs
```diff
@@ -1,11 +1,19 @@
...
+/// Counts the number of set bits in an unsigned integer.
+///
+/// # Arguments
+///
+/// * `n` - An unsigned 32-bit integer whose set bits will be counted.
+///
+/// # Returns
+///
+/// * `usize` - The number of set bits (1s) in the binary representation of the input number.
+pub fn count_set_bits(mut n: usize) -> usize {
```
vil02, line 16: Do you think it makes sense to _remove_ the `mut` here? I do not have a strong opinion.

vil02 (changes requested):

src/bit_manipulation/sum_of_two_integers.rs
```diff
@@ -23,26 +26,29 @@ pub fn add_two_integers(a: i32, b: i32) -> i32 {
...
-        assert_eq!(add_two_integers(3, 5), 8);
-        assert_eq!(add_two_integers(100, 200), 300);
-        assert_eq!(add_two_integers(65535, 1), 65536);
+    macro_rules! test_add_two_integers {
+        ($($name:ident: $test_case:expr,)*) => {
+            $(
+                #[test]
+                fn $name() {
+                    let (a, b, expected) = $test_case;
+                    assert_eq!(add_two_integers(a, b), expected);
```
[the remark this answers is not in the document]
sozelfist: It just work well.

src/bit_manipulation/counting_bits.rs
sozelfist: It's possible to remove `mut` from the function params if we want to maintain a more functional style, but It would require creating a new local variable instead.

In this case, seeing the parameter being mutated directly rather than creating a new variable is more readable.

Honestly, I think using `mut` is the most straightforward and efficient choice here. It doesn't add complexity, and it's a natural fit for the algorithm. If this were part of a larger system with stricter immutability preferences, you might consider avoiding `mut`, but for a utility function like this, it works well as-is.

vil02 (approved):

PR #802, opened 2024-10-02
Refactor Disjoint Set Union

sozelfist:
## Changes

- replaced `reserve_exact()` with `Vec::with_capacity()`, a more idiomatic approach for reserving memory for the nodes vector.
- add doccumentations
- rewrite tests

## Checklist:

### DIFF
--- a/src/graph/disjoint_set_union.rs
+++ b/src/graph/disjoint_set_union.rs
@@ -1,95 +1,148 @@
+//! This module implements the Disjoint Set Union (DSU), also known as Union-Find,
+//! which is an efficient data structure for keeping track of a set of elements
+//! partitioned into disjoint (non-overlapping) subsets.
+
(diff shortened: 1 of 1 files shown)


src/graph/disjoint_set_union.rs
```diff
@@ -1,95 +1,148 @@
...
+    pub fn new(num_elements: usize) -> DisjointSetUnion {
+        let mut nodes = Vec::with_capacity(num_elements + 1);
+        for idx in 0..=num_elements {
+            nodes.push(DSUNode {
+                parent: idx,
+                size: 1,
+            });
         }
+
         DisjointSetUnion { nodes }
```
vil02, line 41:
```suggestion
        Self { nodes }
```

vil02 (approved):

PR #805, opened 2024-10-04
k-d tree (A k-dimensional tree)

bomenderick:
## Description
A K-D Tree(also known as a K-Dimensional Tree) is a binary search tree where data in each node is a K-Dimensional point in space. In short, it is a space partitioning data structure for organizing points in a K-Dimensional space in other to facilitate nearest neighbor search of points.
In addition to insert, search, and delete methods, the implementation also supports nearest neighbors search, median finding for insertion in other to keep the k-d tree balanced, and a merge method to combine two k-d trees by collecting their points and building a balanced k-d tree from it.

Read more: 
- https://www.geeksforgeeks.org/search-and-insertion-in-k-dimensional-tree/
- https://www.geeksforgeeks.org/deletion-in-k-dimensional-tree/


src/data_structures/kd_tree.rs
```diff
@@ -0,0 +1,470 @@
...
+
+impl<T: PartialOrd + Copy, const K: usize> Default for KDTree<T, K> {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+impl<T: PartialOrd + Copy, const K: usize> KDTree<T, K> {
+    // Create and empty kd-tree
+    // #[must_use]
```
vil02, line 39:
```suggestion
```

src/data_structures/kd_tree.rs
```diff
@@ -0,0 +1,470 @@
...
+        KDTree {
+            root: None,
+            size: 0,
+        }
+    }
+
+    // Returns true if point found, false otherwise
+    pub fn contains(&self, point: &[T; K]) -> bool {
+        search_rec(&self.root, point, 0)
+    }
```
vil02, line 50: Is this really needed?

src/data_structures/kd_tree.rs
```diff
@@ -0,0 +1,470 @@
...
+
+    // Returns true if successfully delete a point, false otherwise
```
vil02, line 52: delete?

src/data_structures/kd_tree.rs
```diff
@@ -0,0 +1,470 @@
...
+    where
+        T: Sum + Signed + Real,
+    {
+        let mut neighbors: Vec<(T, [T; K])> = Vec::new();
+        n_nearest_neighbors(&self.root, point, n, 0, &mut neighbors);
+        neighbors
+    }
+
+    // Returns the number of points in a kd-tree
+    // #[must_use]
```
vil02, line 81:
```suggestion
```

src/data_structures/kd_tree.rs
```diff
@@ -0,0 +1,470 @@
...
+    pub fn len(&self) -> usize {
+        self.size
+    }
+
+    // Returns the depth a kd-tree
+    // #[must_use]
```
vil02, line 87:
```suggestion
```

src/data_structures/kd_tree.rs
```diff
@@ -0,0 +1,470 @@
...
+    pub fn depth(&self) -> usize {
+        depth_rec(&self.root, 0, 0)
+    }
```
vil02, line 90: Is this needed?

src/data_structures/kd_tree.rs
```diff
@@ -0,0 +1,470 @@
...
+
+    // Determine whether there exist points in a kd-tree or not
+    // #[must_use]
```
vil02, line 93:
```suggestion
```

src/data_structures/kd_tree.rs
```diff
@@ -0,0 +1,470 @@
...
+    pub fn is_empty(&self) -> bool {
+        self.root.is_none()
+    }
+
+    // Returns a kd-tree built from a vector points
+    // #[must_use]
```
vil02, line 99:
```suggestion
```

src/data_structures/kd_tree.rs
```diff
@@ -0,0 +1,470 @@
...
+            tree.size = points.len();
+            tree.root = build_rec(points, 0);
+
+            tree
+        }
+    }
+
+    /// Returns a `KDTree` containing both trees
+    /// Merging two KDTrees by collecting points and rebuilding
+    // #[must_use]
```
vil02, line 114:
```suggestion
```

src/data_structures/kd_tree.rs
```diff
@@ -0,0 +1,470 @@
...
+    pub fn merge(&mut self, other: &mut Self) -> Self {
+        let mut points: Vec<[T; K]> = Vec::new();
+        collect_points(&self.root, &mut points);
+        collect_points(&other.root, &mut points);
+        KDTree::build(points)
+    }
```
vil02, line 120: Is this really needed?

vil02 (changes requested):
@bomenderick: thanks for interesting contribution.

Before this will be merged some work needs to be done. First of all: please add missing tests. Especially the functionality finding the closest point has to be exercised quite brutally.

Could you also describe/document in the code, which operations does your implementation support?

src/data_structures/kd_tree.rs
bomenderick: `Contains` is useful to search for the presence of a point in a k-d tree.
Or do you mean the additional call of `search_rec`?

vil02: Does `contains` need to be public?

bomenderick: **Yes** for interaction with the outside world. However, `contains` is just to complete `CRUD` operations on a k-d tree.
Indeed, the current k-d tree implementation doesn't make use of `contains`. However, its implementation could help as it is just a synonym of the `search` method of a k-d tree.
Would you recommend I make it private for now or remove it?

PR #809, opened 2024-10-06
feat: add fermats little theorem

saahil-mahato:
### Fermat Primality Test Implementation

This PR introduces an implementation of the **Fermat Primality Test** in Rust. The algorithm leverages Fermat's Little Theorem, which states that if \( p \) is a prime number, then for any integer \( a \) such that \( 1 < a < p - 1 \), it holds that:

a^(p−1) ≡ 1 (mod p)

**Key Features:**
- **Modular Exponentiation:** Efficient computation of \( a^{(p-1)} \mod p \) is performed using a helper function to handle large numbers without overflow.
- **Probabilistic Approach:** The algorithm allows for multiple iterations (`k` tests) with randomly selected bases to increase the reliability of the primality check.
- **Composite Detection:** While it identifies prime numbers effectively, the implementation also recognizes that some composite numbers, known as Carmichael numbers, can pass the test under certain conditions.

**Test Cases:**
- The PR includes comprehensive unit tests for various scenarios, including known prime numbers, composite numbers, small integers, large numbers, and special cases like Carmichael numbers.

This implementation provides a fast, probabilistic approach to primality testing, making it suitable for applications in cryptography and numerical analysis.

## Type of change

## Checklist:

### DIFF
--- a/src/math/fermats_little_theorem.rs
+++ b/src/math/fermats_little_theorem.rs
@@ -1,0 +1,134 @@
+use rand::Rng;
+
+/// Performs modular exponentiation.
+///
+/// This function computes `(base ^ exp) mod modulus` efficiently using the method
+/// of exponentiation by squaring.
+///
+/// # Parameters
+///
+/// - `base`: The base value (u64).
+/// - `exp`: The exponent (u64).
+/// - `modulus`: The modulus (u64).
+///
+/// # Returns
+///
+/// The result of `(base ^ exp) mod modulus`.
+fn mod_exp(base: u64, exp: u64, modulus: u64) -> u64 {
+    let mut result = 1;
+    let mut base = base % modulus;
+    let mut exp = exp;
+
+    while exp > 0 {
+        if exp % 2 == 1 {
+            result = (result * base) % modulus;
+        }
+        exp /= 2;
+        base = (base * base) % modulus;
+    }
+
+    result
+}
+
+/// Implements Fermat's Little Theorem for probabilistic primality testing.
+///
+/// Fermat's Little Theorem states that if `p` is a prime number, then for any integer
+/// `a` such that `1 < a < p`, it holds that:
+///
+/// ```text
+/// a^(p-1) ≡ 1 (mod p)
+/// ```
+///
+/// This function tests if the given number `p` is prime by selecting `k` random integers `a`
+/// in the range `[2, p-1]` and checking if the above condition holds. If it fails for any `a`,
+/// the number is classified as composite. However, if it passes for all chosen values, it is
+/// considered likely prime.
+///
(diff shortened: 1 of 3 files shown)


src/math/fermats_little_theorem.rs
```diff
@@ -0,0 +1,134 @@
...
+    while exp > 0 {
+        if exp % 2 == 1 {
+            result = (result * base) % modulus;
+        }
+        exp /= 2;
+        base = (base * base) % modulus;
+    }
+
+    result
+}
```
vil02, line 31: Why not to use [`modular_exponential`](https://github.com/TheAlgorithms/Rust/blob/0dbaff560587087363a704694895bf310d269197/src/math/modular_exponential.rs#L60)?

src/math/fermats_little_theorem.rs
```diff
@@ -0,0 +1,134 @@
...
+    }
+
+    #[test]
+    fn test_carmichael_numbers() {
+        let carmichael_numbers = vec![561, 1105, 1729, 2465, 2821, 6601];
+        for &n in &carmichael_numbers {
+            fermats_little_theorem(n, 10);
+            // Skip assertion for carmichael numbers
+        }
+    }
```
vil02, line 133: Could you express these tests as [here](https://github.com/TheAlgorithms/Rust/blob/0dbaff560587087363a704694895bf310d269197/src/string/rabin_karp.rs#L89)?

src/math/fermats_little_theorem.rs
```diff
@@ -0,0 +1,134 @@
...
+        assert!(fermats_little_theorem(104729, 10));
+        assert!(!fermats_little_theorem(104730, 10));
+    }
+
+    #[test]
+    fn test_carmichael_numbers() {
+        let carmichael_numbers = vec![561, 1105, 1729, 2465, 2821, 6601];
+        for &n in &carmichael_numbers {
+            fermats_little_theorem(n, 10);
+            // Skip assertion for carmichael numbers
```
vil02, line 131: Why do you skip the assertion here?

vil02 (changes requested):

src/math/fermats_little_theorem.rs
saahil-mahato: Done

src/math/fermats_little_theorem.rs
saahil-mahato: Its because carmichael numbers can be true or false as this is a probabilistic test. I will use a constant seed so that tests don't fail.

saahil-mahato:
@vil02 I have resolved the comments. Please review and let me know if anything is missing.

PR #824, opened 2024-10-17
Cleanup Saddleback Search

sozelfist:
## Checklist:


src/searching/saddleback_search.rs
```diff
@@ -1,85 +1,121 @@
...
+/// corner of the matrix and moves left or down based on comparisons with the target element.
+///
+/// # Arguments
+///
+/// * `matrix` - A 2D vector representing the sorted matrix.
+/// * `element` - The target element to search for.
+///
+/// # Returns
+///
+/// Returns a tuple (row, column) where both indices are 1-based. If the element is not found, returns (0, 0).
```
vil02, line 16: Why no to return `Option<(usize, usize)>`?

vil02 (changes requested):

src/searching/saddleback_search.rs
sozelfist: Return `Option<(usize, usize)>` and 0-based index is more idiomatic.

src/searching/saddleback_search.rs
```diff
@@ -1,85 +1,147 @@
...
+///
+/// # Arguments
+///
+/// * `matrix` - A 2D vector representing the sorted matrix.
+/// * `element` - The target element to search for.
+///
+/// # Returns
+///
+/// Returns `Some((row, column))` where both indices are 0-based. If the element is not found, returns `None`.
+pub fn saddleback_search(matrix: &[Vec<isize>], element: isize) -> Option<(usize, usize)> {
```
vil02, line 17: What if the input is not a proper matrix? (i.e. not a _rectangle_).

src/searching/saddleback_search.rs
```diff
@@ -1,85 +1,147 @@
...
+            2,
+            Some((1, 0)),
+        ),
+        test_single_element: (
+            vec![
+                vec![1],
+            ],
+            1,
+            Some((0, 0)),
+        ),
```
vil02, line 145: Such test case fails:
```suggestion
        ),
        test_empty_rows: (
            vec![
                vec![],
                vec![],
                vec![],
                vec![],
            ],
            1,
            None::<(usize, usize)>,
        ),
```

vil02 (changes requested):

src/searching/saddleback_search.rs
sozelfist: In the case of a proper matrix, we can return an error indicating that the matrix is not rectangular.

src/searching/saddleback_search.rs
sozelfist: I see that, the current implementation does not handle this case. In cases of an empty matrix, or matrix with all empty rows, returning `None` is the best option.

sozelfist:
The added edge cases are just awesome.

sozelfist:
Can you have a look, @vil02?

src/searching/saddleback_search.rs
```diff
@@ -1,85 +1,196 @@
...
-// top-right corner of the matrix and moving left or down based on the comparison of the current
-// element with the target element.
 use std::cmp::Ordering;
 
-pub fn saddleback_search(matrix: &[Vec<i32>], element: i32) -> (usize, usize) {
-    // Initialize left and right indices
+/// Custom error type to represent errors related to matrix validation.
+#[derive(Debug, PartialEq, Eq)]
+pub enum MatrixError {
+    NonRectangularMatrix,
```
vil02, line 6: This is like super nitpicking, but can matrix be non rectangular? Maybe:
```suggestion
    NonRectangularInput,
```
would be better?

src/searching/saddleback_search.rs
```diff
@@ -1,85 +1,196 @@
...
+
+/// Performs Saddleback search on a sorted 2D matrix.
+///
+/// The Saddleback search algorithm finds the position of a target element in a matrix where
+/// each row and each column is sorted in ascending order. The search starts from the top-right
+/// corner of the matrix and moves left or down based on comparisons with the target element.
+///
+/// # Arguments
+///
+/// * `matrix` - A 2D vector representing the sorted matrix.
```
vil02, line 17: 2D vector is a bit misleading here. Maybe an expression like _vector of vectors_ would be better? Or _2D array_?

src/searching/saddleback_search.rs
```diff
@@ -1,85 +1,196 @@
...
 
-pub fn saddleback_search(matrix: &[Vec<i32>], element: i32) -> (usize, usize) {
-    // Initialize left and right indices
+/// Custom error type to represent errors related to matrix validation.
+#[derive(Debug, PartialEq, Eq)]
+pub enum MatrixError {
+    NonRectangularMatrix,
+}
+
+/// Performs Saddleback search on a sorted 2D matrix.
```
vil02, line 9: What do you think about introducing a `is_sorted` function, checking if the matrix is sorted as expected. Clearly if it would be used inside the `saddleback_search` function it would spoil the complexity, but if would be great if the user could verify if their matrix is as expected. I guess the possible workflow could look like this:
- the used generates some data,
- check the data with `is_sorted`,
- uses many times the `saddleback_search`.

Having such function would also _define_ what does it mean to have a _sorted matrix_.

vil02 (changes requested):

src/searching/saddleback_search.rs
sozelfist: Yes, It's a good idea.

src/searching/saddleback_search.rs
sozelfist: Referring to `matrix` as a _2D vector_ can be a bit confusing, especially since Rust doesn't have native 2D arrays like some other languages. In Rust, the type we are working with is a vector of vectors (`Vec<Vec<isize>>`), which represents a matrix-like structure but isn't truly a 2D array.

src/searching/saddleback_search.rs
sozelfist: Adding an `is_sorted` function is a great idea. It allows users to validate whether their matrix is correctly sorted before running the `saddleback_search` and provides a clear definition of a _sorted matrix_. It can be useful when users generate their data or when matrices are passed from external sources to ensure the structure is valid.

src/searching/saddleback_search.rs
sozelfist, line 6:
```suggestion
    NonRectangularMatrix,
    NotSorted
```

Add new error type for unsorted matrix

src/searching/saddleback_search.rs
```diff
@@ -1,85 +1,196 @@
...
 use std::cmp::Ordering;
 
-pub fn saddleback_search(matrix: &[Vec<i32>], element: i32) -> (usize, usize) {
-    // Initialize left and right indices
+/// Custom error type to represent errors related to matrix validation.
+#[derive(Debug, PartialEq, Eq)]
+pub enum MatrixError {
+    NonRectangularMatrix,
+}
+
```
sozelfist, line 8: I propose the `is_sorted` function as follow, do you have any suggestion, @vil02?

```suggestion

/// Checks if the given matrix (vector of vectors) is sorted row-wise and column-wise.
///
/// A matrix is considered sorted if:
///
/// * Each row is sorted in non-decreasing order.
/// * Each column is sorted in non-decreasing order.
///
/// # Arguments
///
/// * `matrix` - A vector of vectors representing the matrix to check.
///
/// # Returns
///
/// Returns `true` if the matrix is sorted both row-wise and column-wise. Otherwise, returns `false`.
pub fn is_sorted(matrix: &[Vec<isize>]) -> bool {
    if matrix.is_empty() || matrix.iter().all(|row| row.is_empty()) {
        return true;
    }

    let rows = matrix.len();
    let cols = matrix[0].len();

    // Check if rows are sorted.
    for row in matrix {
        if row.windows(2).any(|w| w[0] > w[1]) {
            return false;
        }
    }

    // Check if columns are sorted.
    for col in 0..cols {
        for row in 1..rows {
            if matrix[row - 1][col] > matrix[row][col] {
                return false;
            }
        }
    }

    true
}

```

src/searching/saddleback_search.rs
```diff
@@ -1,85 +1,196 @@
...
+///
+/// # Returns
+///
+/// Returns `Ok(Some((row, column)))` where both indices are 0-based if the element is found.
+/// Returns `Ok(None)` if the element is not found.
+/// Returns `Err(MatrixError)` if the matrix is not rectangular.
+pub fn saddleback_search(
+    matrix: &[Vec<isize>],
+    element: isize,
+) -> Result<Option<(usize, usize)>, MatrixError> {
```
sozelfist, line 28: Then, we can check if the input was sorted as follow:

```suggestion
pub fn saddleback_search(
    matrix: &[Vec<isize>],
    element: isize,
    check_sorted: bool,
) -> Result<Option<(usize, usize)>, MatrixError> {
```

src/searching/saddleback_search.rs
```diff
@@ -1,85 +1,196 @@
...
+    if matrix.is_empty() || matrix.iter().all(|row| row.is_empty()) {
+        return Ok(None);
+    }
+
+    if matrix.iter().any(|row| row.len() != matrix[0].len()) {
+        return Err(MatrixError::NonRectangularMatrix);
+    }
+
```
sozelfist, line 36:
```suggestion
    if check_sorted && !is_sorted(matrix) {
        return Err(MatrixError::NotSorted);
    }
```

src/searching/saddleback_search.rs
```diff
@@ -1,85 +1,196 @@
...
+        test_full_empty_rows: (
+            vec![
+                vec![],
+                vec![],
+                vec![],
+                vec![],
+            ],
+            1,
+            Ok(None::<(usize, usize)>),
+        ),
```
sozelfist, line 194:
```suggestion
        ),
        test_unsorted_matrix_with_check: (
            vec![
                vec![1, 10, 100],
                vec![20, 200, 2],
                vec![3, 30, 300],
            ],
            200,
            Err::<Option<(usize, usize)>, MatrixError>(MatrixError::NotSorted),
        ),
```

src/searching/saddleback_search.rs
```diff
@@ -1,85 +1,255 @@
...
+/// Checks if the given matrix (vector of vectors) is sorted row-wise and column-wise.
+///
+/// # Arguments
+///
+/// * `matrix` - A vector of vectors representing the matrix to check.
+///
+/// # Returns
+///
+/// Returns `true` if the matrix is sorted both row-wise and column-wise. Otherwise, returns `false`.
+fn is_sorted(matrix: &[Vec<isize>]) -> bool {
```
vil02, line 19:
```suggestion
pub fn is_sorted(matrix: &[Vec<isize>]) -> bool {
```

Imagine that you have some matrix. You know that you will query the position of many elements and you want to check if the matrix is soared only once.

src/searching/saddleback_search.rs
```diff
@@ -1,85 +1,255 @@
...
+    let rows = matrix.len();
+    let cols = matrix[0].len();
```
vil02, line 21: I think we need to check if the input is _rectangular_.

vil02 (changes requested):

src/searching/saddleback_search.rs
sozelfist: We can move the logic that checks the matrix form and `is_sorted` into a single function, for example `validate_matrix`

src/searching/saddleback_search.rs
sozelfist: I understand this point. By the way, do you know how can we express the tests here? Suppose, we have two functions, one for validate matrix, one for saddleback search like this:
```rs
fn validate_matrix(matrix: &[Vec<isize>]) -> Result<(), MatrixError> {
    // Check if all rows have the same length (rectangular matrix).
    if matrix.iter().any(|row| row.len() != matrix[0].len()) {
        return Err(MatrixError::NonRectangularInput);
    }

    // Check if matrix is sorted.
    if !is_sorted(matrix) {
        return Err(MatrixError::NotSorted);
    }

    Ok(())
}
```
and
```rs
fn saddleback_search(matrix: &[Vec<isize>], element: isize) -> Option<(usize, usize)> {
    if matrix.is_empty() || matrix.iter().all(|row| row.is_empty()) {
        return None;
    }

    let mut left_index = 0;
    let mut right_index = matrix[0].len() - 1;

    while left_index < matrix.len() {
        match element.cmp(&matrix[left_index][right_index]) {
            Ordering::Equal => return Some((left_index, right_index)),
            Ordering::Greater => {
                left_index += 1;
            }
            Ordering::Less => {
                if right_index == 0 {
                    break;
                } else {
                    right_index -= 1;
                }
            }
        }
    }

    None
}
```

Can you propose how we can write tests, @vil02?

src/searching/saddleback_search.rs
vil02: As above: I would introduce a new function checking if the input is _rectangular_ and use it in both public functions.

src/searching/saddleback_search.rs
sozelfist: Can you suggest the detailed changes?

src/searching/saddleback_search.rs
```diff
@@ -1,85 +1,255 @@
-// Saddleback search is a technique used to find an element in a sorted 2D matrix in O(m + n) time,
-// where m is the number of rows, and n is the number of columns. It works by starting from the
-// top-right corner of the matrix and moving left or down based on the comparison of the current
-// element with the target element.
 use std::cmp::Ordering;
 
-pub fn saddleback_search(matrix: &[Vec<i32>], element: i32) -> (usize, usize) {
```
Nonmembersshit, line 7: n o it will cquse crash

sozelfist: Please do not spam here.

PR #825, opened 2024-10-17
Refactor Anagram

sozelfist:
## Checklist:


sozelfist:
Can you have a look, @vil02?

src/string/anagram.rs
```diff
@@ -1,27 +1,101 @@
...
+/// Computes the frequency of characters in a string.
+///
+/// # Arguments
+///
+/// * `s` - Input string.
+///
+/// # Returns
+///
+/// * A `HashMap` where the keys are characters and values are their frequencies.
+fn char_frequency(s: &str) -> HashMap<char, usize> {
```
vil02, line 61:
```suggestion
fn char_count(s: &str) -> HashMap<char, usize> {
```
It is not the _frequency_.

vil02 (changes requested):

src/string/anagram.rs
sozelfist: Will rename the function

src/string/anagram.rs
```diff
@@ -1,27 +1,111 @@
...
+///
+/// # Returns
+///
+/// * A `HashMap` where the keys are characters and values are their frequencies.
+fn char_count(s: &str) -> HashMap<char, usize> {
+    let mut freq = HashMap::new();
+    for c in s.chars() {
+        *freq.entry(c).or_insert(0) += 1;
+    }
+    freq
```
vil02, line 66:
```suggestion
/// * A `HashMap` where the keys are characters and values are their count.
fn char_count(s: &str) -> HashMap<char, usize> {
    let mut res = HashMap::new();
    for c in s.chars() {
        *res.entry(c).or_insert(0) += 1;
    }
    res
```

src/string/anagram.rs
```diff
@@ -1,27 +1,111 @@
...
+            if c.is_alphabetic() {
+                Ok(c.to_ascii_lowercase())
+            } else {
+                Err(AnagramError::NonAlphabeticCharacter)
+            }
+        })
+        .collect()
+}
+
+/// Computes the frequency of characters in a string.
```
vil02, line 52:
```suggestion
/// Computes the histogram of characters in a string.
```

vil02 (approved):

PR #829, opened 2024-10-20
create financial/price_plus_tax function

alirezasariri78:
# Pull Request Template

## Description

## Type of change

## Checklist:


src/financial/README.md
vil02, line 1: This file is not needed. Please remove it.

src/financial/price_plus_tax.rs
vil02, line 1: This seems to be _too trivial_. I am not sure if it is needed.

src/financial/present_value.rs
```diff
@@ -0,0 +1,69 @@
...
+        );
+    }
+
+    #[test]
+    fn test_present_value_zero_discount_rate() {
+        assert_eq!(
+            184924.55,
+            present_value(0.0, vec![109129.39, 30923.23, 15098.93, 29734.0, 39.0]).unwrap()
+        );
+    }
```
vil02, line 68: Could you please express these tests using _macros_, cf. [`shortest_palindrome.rs`](https://github.com/TheAlgorithms/Rust/blob/c906255c51a757d15edbffe2253b8b4225a76a2a/src/string/shortest_palindrome.rs).

src/financial/present_value.rs
```diff
@@ -0,0 +1,69 @@
...
+        return Err(PresentValueError::EmptyCashFlow);
+    }
+
+    let present_value = cash_flows
+        .iter()
+        .enumerate()
+        .map(|(i, &cash_flow)| cash_flow / (1.0 + discount_rate).powi(i as i32))
+        .sum::<f64>();
+
+    Ok((present_value * 100.0).round() / 100.0)
```
vil02, line 21: Consider creating a separate function doing the rounding. It has to be tested.

vil02 (changes requested):

src/financial/price_plus_tax.rs
alirezasariri78: based on this link in python repository i add this module :
https://github.com/TheAlgorithms/Python/blob/master/financial/price_plus_tax.py

alirezasariri78:
changes are done as you said :)

src/financial/present_value.rs
alirezasariri78: that is nice idea , ill do it

src/financial/present_value.rs
vil02, line 1: Could you please document _here_ what does this file provide?

vil02 (changes requested):

vil02:
@alirezasariri78: please make sure to update your branch - there are some new warnings in clippy and without the update the CI will fail (cf. #838).

alirezasariri78:
document added to module file

vil02 (approved):

PR #830, opened 2024-10-20
Length conversion

ali77gh:
# Length Conversions

## Description

Algorithm of converting different length units together.

Something like [python repo](https://github.com/TheAlgorithms/Python/blob/master/conversions/length_conversion.py)

 Available Units:
 - Millimeter
 - Centimeter
 - Meter
 - Kilometer
 - Inch
 - Foot
 - Yard
 - Mile

## Type of change

## Checklist:


src/conversions/length_conversion.rs
```diff
@@ -0,0 +1,107 @@
...
+        assert_eq!(length_conversion(4.0, Yard, Kilometer), 0.0036576);
+        assert_eq!(length_conversion(3.0, Foot, Inch), 36.00000000000001);
+        assert_eq!(length_conversion(4.0, Mile, Kilometer), 6.43736);
+        assert_eq!(length_conversion(2.0, Mile, Inch), 126719.68503937007);
+        assert_eq!(length_conversion(3.0, Millimeter, Centimeter), 0.3);
+        assert_eq!(
+            length_conversion(3.0, Millimeter, Inch),
+            0.11811023622047245
+        );
+    }
```
vil02, line 106: Could you suggest some other testing strategy? Imagine that one wants to introduce a new unit... it is not clear how to test it.

The nature of the problem is _linear_, so it should be enough to check each conversion for _two_ lengths/points.

In each unit, the length of $0$ is $0$ - that would be first point.

The second one could be the length of $1$ meter. What do you think about creating a dictionary/map, where the key is the unit, and the value is the length of $1$ meter in that unit? The test could just iterate thought that map and check all the possibilities.

src/conversions/length_conversion.rs
```diff
@@ -0,0 +1,107 @@
...
+
+    /// This function give you a number (let's call it n)
+    /// So if you multiple a value in meters to n you will get unit
+    ///
+    /// m * meter = in-unit
+    fn get_unit_from_meter_multiplier(&self) -> f64 {
+        1.0 / self.get_unit_to_meter_multiplier()
+    }
+}
```
vil02, line 56: Could we just use non-public functions for that?

vil02 (changes requested):

src/conversions/length_conversion.rs
ali77gh: Hello🖐️!
Thanks for your response.

I did add what you said (If I understand it correctly).

If there is more things to change. just let me know.

src/conversions/length_conversion.rs
ali77gh: Hmm🤔
I don't understand you. 
These functions are non-public (private) 
Isn't it?

If you mean I should make these public. the caller can easily do this:
```rust
length_conversion(value, LengthUnit::Meter, Unit);
// and reverse:
length_conversion(value, Unit, LengthUnit::Meter);
```

Should we have different API just for meter

Or maybe I should move my length_conversion to impl block?

Can you explain a bit more?

src/conversions/length_conversion.rs
```diff
@@ -0,0 +1,139 @@
...
+            (Inch, 39.37007874015748f64),
+            (Foot, 3.280839895013123f64),
+            (Yard, 1.0936132983377078f64),
+            (Mile, 0.0006213727366498068f64),
+        ]);
+
+        for (k, v) in map {
+            assert_eq!(length_conversion(1f64, Meter, k), v);
+        }
+    }
```
vil02, line 138:
```suggestion
    #[test]
    fn length_of_one_meter() {
        let meter_in_different_units = HashMap::from([
            (Millimeter, 1000f64),
            (Centimeter, 100f64),
            (Kilometer, 0.001f64),
            (Inch, 39.37007874015748f64),
            (Foot, 3.280839895013123f64),
            (Yard, 1.0936132983377078f64),
            (Mile, 0.0006213727366498068f64),
        ]);

        for (input_unit, input_value) in &meter_in_different_units {
            for (target_unit, target_value) in &meter_in_different_units {
                assert!(
                    num_traits::abs(
                        length_conversion(*input_value, *input_unit, *target_unit) - *target_value
                    ) < 0.0000001
                );
            }
        }
    }
```
Please notice, that the `hash_map` could be a _variable_.

src/conversions/length_conversion.rs
vil02: The entire implementation could look like this:
```rust
fn unit_to_meter_multiplier(from: LengthUnit) -> f64 {
    match from {
        LengthUnit::Millimeter => 0.001,
        LengthUnit::Centimeter => 0.01,
        LengthUnit::Meter => 1.0,
        LengthUnit::Kilometer => 1000.0,
        LengthUnit::Inch => 0.0254,
        LengthUnit::Foot => 0.3048,
        LengthUnit::Yard => 0.9144,
        LengthUnit::Mile => 1609.34,
    }
}

fn to_meter(input: f64, from: LengthUnit) -> f64 {
    input * unit_to_meter_multiplier(from)
}

fn meter_to_unit(input: f64, to: LengthUnit) -> f64 {
    input / unit_to_meter_multiplier(to)
}

pub fn length_conversion(input: f64, from: LengthUnit, to: LengthUnit) -> f64 {
    meter_to_unit(to_meter(input, from), to)
}
```

In my opinion, it is easier to think about _conversion to meter_ and _conversion from meter_ than about multipliers.

src/conversions/length_conversion.rs
```diff
@@ -0,0 +1,139 @@
...
+        assert_eq!(length_conversion(4.0, Yard, Kilometer), 0.0036576);
+        assert_eq!(length_conversion(3.0, Foot, Inch), 36.00000000000001);
+        assert_eq!(length_conversion(4.0, Mile, Kilometer), 6.43736);
+        assert_eq!(length_conversion(2.0, Mile, Inch), 126719.68503937007);
+        assert_eq!(length_conversion(3.0, Millimeter, Centimeter), 0.3);
+        assert_eq!(
+            length_conversion(3.0, Millimeter, Inch),
+            0.11811023622047245
+        );
+    }
```
vil02, line 108:
```suggestion
```
I think this is not needed.

src/conversions/length_conversion.rs
```diff
@@ -0,0 +1,139 @@
...
+        let units = vec![
+            Millimeter, Centimeter, Meter, Kilometer, Inch, Foot, Yard, Mile,
+        ];
+
+        for u1 in units.clone() {
+            for u2 in units.clone() {
+                assert_eq!(length_conversion(0f64, u1, u2), 0f64);
+            }
+        }
+    }
```
vil02, line 121: What do you think about using a similar approach as in `length_of_one_meter`?

vil02 (changes requested):

src/conversions/length_conversion.rs
ali77gh: I agree. It's cleaner👍.
I did change it 
https://github.com/ali77gh/TheAlgorithmsRust/commit/ea60fed02ae820e8968b6c8ce8e7c205cd6f66b1.

src/conversions/length_conversion.rs
ali77gh: Ok!
Its here:
https://github.com/ali77gh/TheAlgorithmsRust/commit/ea60fed02ae820e8968b6c8ce8e7c205cd6f66b1

src/conversions/length_conversion.rs
ali77gh: I removed these:
https://github.com/ali77gh/TheAlgorithmsRust/commit/ea60fed02ae820e8968b6c8ce8e7c205cd6f66b1

src/conversions/length_conversion.rs
```diff
@@ -0,0 +1,104 @@
...
+
+/// Private methods
+/// This is a n*n problem
+/// It's gonna be 56 functions or 56 cases in match statement
+/// It's hard to write code for every unit to unit so I solve this problem
+/// by converting input to meter and than convert it to output
+/// This function give you a number (let's call it n)
+/// So if you multiple a value in this unit to n you will get meters
+///
+/// m * in-unit = meter
```
vil02, line 35:
```suggestion
```

src/conversions/length_conversion.rs
```diff
@@ -0,0 +1,104 @@
...
+/// -> Wikipedia reference: https://en.wikipedia.org/wiki/Millimeter
+/// -> Wikipedia reference: https://en.wikipedia.org/wiki/Centimeter
+/// -> Wikipedia reference: https://en.wikipedia.org/wiki/Meter
+/// -> Wikipedia reference: https://en.wikipedia.org/wiki/Kilometer
+/// -> Wikipedia reference: https://en.wikipedia.org/wiki/Inch
+/// -> Wikipedia reference: https://en.wikipedia.org/wiki/Foot
+/// -> Wikipedia reference: https://en.wikipedia.org/wiki/Yard
+/// -> Wikipedia reference: https://en.wikipedia.org/wiki/Mile
+
+/// Universal Units on Length
```
vil02, line 14:
```suggestion
```

vil02 (changes requested):

vil02 (approved):

PR #840, opened 2024-11-06
add electronics/coulombs law module

alirezasariri78:
# Pull Request Template

## Description

## Type of change

## Checklist:


src/matrix/search_in_a_sorted_matrix.rs
vil02, line 1: Is this file relevant in this PR? If not, please remove it.

src/electronics/coulombs_law.rs
```diff
@@ -0,0 +1,183 @@
...
+        return Ok(format!(
+            "charge2: {}",
+            calculate_charge(charge1, force, distance)
+        ));
+    }
+    Ok(format!(
+        "distance: {}",
+        calculate_distance(charge_product, force)
+    ))
+}
```
vil02, line 58: Is this function really needed? If you insist on having it, please consider using `Optional` as inputs.

vil02 (changes requested):

src/matrix/search_in_a_sorted_matrix.rs
alirezasariri78: this module is located in python repository : https://github.com/TheAlgorithms/Python/blob/master/matrix/searching_in_sorted_matrix.py

src/electronics/coulombs_law.rs
alirezasariri78: based on electronic modules in python repository : https://github.com/TheAlgorithms/Python/blob/master/electronics/coulombs_law.py

alirezasariri78: using optional would not cause boiler plate ?
in this situation i should double check value of inputs (none or some and sign of inputs)

PR #861, opened 2025-01-26
Addition of multiple_longest_common_subsequence

Gaspardcode:
## Description

To whom it may concern,

this pull request aims at introducing an algorithm to find the longest common subsequence among multiple sequences. 

Finding a solution other than brute force is not really intuitive. I advise to look at the first comments of [this thread](https://codeforces.com/blog/entry/3188).

The pull request implements a method based on heuristic, as described in the following [research paper](https://github.com/epita-rs/MLCS/blob/main/paper.pdf).

Thank you for considering this work.

## Type of change

## Checklist:

Note: I mean I did only delete the issues related to my pull request


Gaspardcode:
Hello, 

I might need help with this pull request, especially `cargo clippy --all -- -D warnings`.

Surely I am not supposed to fix the 500+ errors already present before, right ?
I tried to run `cargo clippy --fix` but it wouldn't do.

 Thank you

Gaspardcode:
image

vil02:
> Hello,
> 
> I might need help with this pull request, especially `cargo clippy --all -- -D warnings`.
> 
> Surely I am not supposed to fix the 500+ errors already present before, right ? I tried to run `cargo clippy --fix` but it wouldn't do.
> 
> Thank you

@Gaspardcode: Try updating your branch. There were some new warnings introduced, but it should work now  (cf. #871).

Gaspardcode:
> > Hello,
> > I might need help with this pull request, especially `cargo clippy --all -- -D warnings`.
> > Surely I am not supposed to fix the 500+ errors already present before, right ? I tried to run `cargo clippy --fix` but it wouldn't do.
> > Thank you
> 
> @Gaspardcode: Try updating your branch. There were some new warnings introduced, but it should work now (cf. #871).

Hello, I did indeed fixed it by simply pulling, thank you

src/string/multiple_longest_common_subsequence.rs
```diff
@@ -0,0 +1,492 @@
+use std::cmp::max;
+use std::cmp::Ordering;
+use std::collections::HashMap;
+
+const IMPOSSIBLE_NB: usize = 999_999_999_999;
```
vil02, line 5: Would it be better to define `IMPOSSIBLE_NB` as the maximal value, which could be stored in `usize`?

Or even better, the corresponding vectors should store `Option<usize>`?

src/string/multiple_longest_common_subsequence.rs
```diff
@@ -0,0 +1,492 @@
...
+            successors.push(succ);
+        }
+        // else we discard it and move on to the next letter
+    }
+    successors
+}
+
+/// Computes the heuristic function given a point
+/// min ( { M_ij[ p[i] ][ p[j] ] | (i,j) in [0 ; d] } )
+/// [Documentation](https://github.com/epita-rs/MLCS/blob/main/paper.pdf)
```
vil02, line 165:
```suggestion
/// [Documentation](https://github.com/epita-rs/MLCS/blob/main/doc/paper.pdf)
```

src/string/multiple_longest_common_subsequence.rs
```diff
@@ -0,0 +1,492 @@
...
+            }
+        }
+    }
+
+    *similarity.iter().min().unwrap()
+}
+
+/// Add the first matches to the queue
+/// For each starting point found, sets an impossible point as parent
+/// [Documentation](https://github.com/epita-rs/MLCS/blob/main/paper.pdf)
```
vil02, line 181:
```suggestion
/// [Documentation](https://github.com/epita-rs/MLCS/blob/main/doc/paper.pdf)
```

src/string/multiple_longest_common_subsequence.rs
```diff
@@ -0,0 +1,492 @@
...
+                };
+            }
+        }
+    }
+
+    matrix
+}
+
+//given given 2D coordinates, translates into 1D coordinates
+fn translate(i: usize, j: usize, d: usize) -> usize {
```
vil02, line 371: What do you think about renaming this function to `to_linear_index`?

src/string/multiple_longest_common_subsequence.rs
```diff
@@ -0,0 +1,492 @@
...
+/// * `String' if no LCS was found
+pub fn multiple_longest_common_subsequence(chains: &Vec<&str>) -> String {
+    const C: u64 = 20;
+
+    // Preprocessing
+    let mut ctx = Context::new(chains);
+
+    // queue
+    let mut queue: Vec<Vec<usize>> = vec![];
+    init_queue(&mut ctx, &mut queue);
```
vil02, line 295: This should look more like this:
```suggestion
    let mut queue: Vec<Vec<usize>> = init_queue(&mut ctx);
```

src/string/multiple_longest_common_subsequence.rs
```diff
@@ -0,0 +1,492 @@
...
+}
+
+//given given 2D coordinates, translates into 1D coordinates
+fn translate(i: usize, j: usize, d: usize) -> usize {
+    i * d + j
+}
+
+// given a point p and his successor q, computes necessary informations
+// point p is marked PARENT of q
+fn update_suc(ctx: &mut Context, p: Vec<usize>, q: Vec<usize>) {
```
vil02, line 377: Should this be a method of `Context`?

vil02 (changes requested):

src/string/multiple_longest_common_subsequence.rs
Gaspardcode: Sure thing, I moved functions with a similar prototype to `Context` methods

src/string/multiple_longest_common_subsequence.rs
Gaspardcode: sorry about that one !

src/string/multiple_longest_common_subsequence.rs
Gaspardcode: Yes it would match better Rust's identity.

NB: since thoses vectors are moved around a lot, I suspect there could be a small overhead caused by frequent unwrapping, but it doesnt matter here.

src/string/multiple_longest_common_subsequence.rs
```diff
@@ -0,0 +1,490 @@
...
+/// * `String` if a Longest Common Subsequence exists
+/// * `String' if no LCS was found
+pub fn multiple_longest_common_subsequence(chains: &Vec<&str>) -> String {
+    const C: u64 = 20;
+
+    // Preprocessing
+    let mut ctx = Context::new(chains);
+
+    // queue
+    let mut queue: Vec<Vec<Option<usize>>> = ctx.init_queue();
```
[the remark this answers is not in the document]
Gaspardcode: Hi, to be honest, I won't have the time to prove you wrong this week, but I've tried, and its actually more complicated than it intuitively feels.
Can't recall much unfortunately, I'd have to dig into it again.

PR #874, opened 2025-03-06
Added move_to_front_encoding implementation

Divakar-2508:
## **Description**  

This PR adds the **Move-To-Front (MTF) Encoding** algorithm under the **compression** category.  

### **Algorithm Overview**  
Move-To-Front (MTF) Encoding is a lossless data compression algorithm that replaces each character with its position in a dynamically updated list. Every time a character is accessed, it moves to the front of the list. This technique is commonly used as a preprocessing step before further entropy encoding, such as Burrows-Wheeler Transform (BWT).  

For more details: [[Move-To-Front Transform - Wikipedia](https://en.wikipedia.org/wiki/Move-to-front_transform)](https://en.wikipedia.org/wiki/Move-to-front_transform)  

### **Implementation Details**  
- Uses a **Vec<u8>** as a lookup queue for efficient front insertions.  
- Implements both **encoding** and **decoding** functions.  
- Includes **unit tests** to verify correctness.  

## **Type of Change**  
- [x] **New feature** (non-breaking change which adds functionality)  

## **Checklist**  

- [x] I ran the following commands using the latest version of **rust nightly**.  
- [x] I checked `CONTRIBUTING.md`, and my code follows its guidelines.

### DIFF
--- a/src/compression/move_to_front.rs
+++ b/src/compression/move_to_front.rs
@@ -1,0 +1,56 @@
+// https://en.wikipedia.org/wiki/Move-to-front_transform
+
+pub fn move_to_front_encode(text: &str) -> Vec<u8> {
+    dbg!(text);
+    let mut char_table: Vec<char> = (0..=255).map(|ch| ch as u8 as char).collect();
+    let mut result = Vec::new();
+
+    for ch in text.chars() {
+        if let Some(position) = char_table.iter().position(|&x| x == ch) {
+            result.push(position as u8);
+            char_table.remove(position);
+            char_table.insert(0, ch);
+        }
+    }
+
+    result
+}
+
+pub fn move_to_front_decode(encoded: &[u8]) -> String {
+    let mut char_table: Vec<char> = (0..=255).map(|ch| ch as u8 as char).collect();
+    let mut result = String::new();
+
+    for &pos in encoded {
+        let ch = char_table[pos as usize];
+        result.push(ch);
+        char_table.remove(pos as usize);
+        char_table.insert(0, ch);
+    }
+
+    result
+}
+
+#[cfg(test)]
+mod test {
+    use super::*;
+
+    #[test]
+    fn test_move_to_front_encode() {
+        assert_eq!(move_to_front_encode(""), []);
+        assert_eq!(move_to_front_encode("@"), [64]);
+        assert_eq!(move_to_front_encode("aaba"), [97, 0, 98, 1]);
+        assert_eq!(move_to_front_encode("aZ!"), [97, 91, 35]);
+        assert_eq!(move_to_front_encode("banana"), [98, 98, 110, 1, 1, 1]);
+        assert_eq!(move_to_front_encode("\0\n\t"), [0, 10, 10]);
+    }
+
+    #[test]
+    fn test_move_to_front_decode() {
+        assert_eq!(move_to_front_decode(&[]), "");
+        assert_eq!(move_to_front_decode(&[64]), "@");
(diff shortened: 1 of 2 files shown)


src/compression/move_to_front.rs
```diff
@@ -0,0 +1,56 @@
+// https://en.wikipedia.org/wiki/Move-to-front_transform
+
+pub fn move_to_front_encode(text: &str) -> Vec<u8> {
+    dbg!(text);
```
vil02, line 4: This is not needed:

```suggestion
```

src/compression/move_to_front.rs
```diff
@@ -0,0 +1,56 @@
...
+    let mut char_table: Vec<char> = (0..=255).map(|ch| ch as u8 as char).collect();
```
vil02, line 5: What do you think about introducing a function like:
```rust
fn blank_char_table() -> Vec<char> {
    (0..=255).map(|ch| ch as u8 as char).collect()
}
```
and then
```suggestion
    let mut char_table: Vec<char> = blank_char_table();
```
and same in line 20.

src/compression/move_to_front.rs
```diff
@@ -0,0 +1,56 @@
...
+            char_table.remove(position);
+            char_table.insert(0, ch);
+        }
+    }
+
+    result
+}
+
+pub fn move_to_front_decode(encoded: &[u8]) -> String {
+    let mut char_table: Vec<char> = (0..=255).map(|ch| ch as u8 as char).collect();
```
vil02, line 20:
```suggestion
    let mut char_table: Vec<char> = blank_char_table();
```

src/compression/move_to_front.rs
```diff
@@ -0,0 +1,56 @@
...
+
+    #[test]
+    fn test_move_to_front_decode() {
+        assert_eq!(move_to_front_decode(&[]), "");
+        assert_eq!(move_to_front_decode(&[64]), "@");
+        assert_eq!(move_to_front_decode(&[97, 0, 98, 1]), "aaba");
+        assert_eq!(move_to_front_decode(&[97, 91, 35]), "aZ!");
+        assert_eq!(move_to_front_decode(&[98, 98, 110, 1, 1, 1]), "banana");
+        assert_eq!(move_to_front_decode(&[0, 10, 10]), "\0\n\t");
+    }
```
vil02, line 55: What do you think about expressing these tests in a more _parametrized_ way:
```suggestion
    macro_rules! test_mtf {
        ($($name:ident: $tc:expr,)*) => {
        $(
            #[test]
            fn $name() {
                let (text, encoded) = $tc;
                assert_eq!(move_to_front_encode(text), encoded);
                assert_eq!(move_to_front_decode(&encoded), text);
            }
        )*
        }
    }

    test_mtf! {
        empty: ("", []),
        single_character: ("@", [64]),
        regular_0: ("aaba", [97, 0, 98, 1]),
        regular_1: ("aZ!", [97, 91, 35]),
        regular_2: ("banana", [98, 98, 110, 1, 1, 1]),
        special_white_chars: ("\0\n\t", [0, 10, 10]),
    }
```

The advantage it, that you do not need to duplicate the test data.

vil02 (changes requested):

src/compression/move_to_front.rs
Divakar-2508: changed the two

src/compression/move_to_front.rs
Divakar-2508: i'm not accustomed to macros for testing, but this looks clean, applied, thanks!

vil02 (approved):

PR #875, opened 2025-03-11
Add `Euler's Totient` function implementation

allansomensi:
Adds the implementation of the **Euler's Totient function** with a unit test.

## Type of change

## Checklist:


src/math/euler_totient.rs
```diff
@@ -0,0 +1,45 @@
+/// ## Euler's Totient (`φ(n)`)
+///
+/// Calculate the **Euler's Totient** function of a given number `n`.
+///
+/// Wikipedia: https://en.wikipedia.org/wiki/Euler%27s_totient_function
+pub fn euler_totient(mut n: i64) -> i64 {
```
vil02, line 6: What do you think about having the following signature:
```suggestion
pub fn euler_totient(n: u64) -> Result<u64, &'static str> {
```

src/math/euler_totient.rs
```diff
@@ -0,0 +1,45 @@
...
+    fn test_euler_totient() {
+        assert_eq!(euler_totient(1), 1);
+        assert_eq!(euler_totient(325), 240);
+        assert_eq!(euler_totient(746), 372);
+        assert_eq!(euler_totient(3_639), 2_424);
+        assert_eq!(euler_totient(98_354), 49_176);
+        assert_eq!(euler_totient(123_456), 41_088);
+        assert_eq!(euler_totient(493_123_235), 347_518_080);
+        assert_eq!(euler_totient(945_243_784_032), 315_074_904_192);
+        assert_eq!(euler_totient(372_036_854_775_808), 185_661_377_740_800);
```
vil02, line 43: What do you think about having more _parametrized_ approach like in [`test_aliquot_sum`](https://github.com/TheAlgorithms/Rust/blob/b4aecf4b1a30626b0aca1fe40b588b3144858e4c/src/math/aliquot_sum.rs#L33).

vil02 (changes requested):
This functionality is (almost) provided by [`compute_totient.rs`](https://github.com/TheAlgorithms/Rust/blame/b4aecf4b1a30626b0aca1fe40b588b3144858e4c/src/number_theory/compute_totient.rs). Both implementations are similarly efficient. @siriak what do you suggest?

allansomensi:
Oops! I hadn’t noticed this algorithm was already here. Thanks for pointing it out! Closing the PR.

PR #881, opened 2025-05-21
Improve normalize in karatsuba_multiplication implementation

dxdw233:
# Pull Request Template

## Description

Refactor `normalize` in `karatsuba_multiplication.rs` to improve performance and readability.

## Type of change

## Checklist:

### DIFF
--- a/src/math/karatsuba_multiplication.rs
+++ b/src/math/karatsuba_multiplication.rs
@@ -35,8 +35,9 @@ fn _multiply(num1: i128, num2: i128) -> i128 {
 }
 
 fn normalize(mut a: String, n: usize) -> String {
-    for (counter, _) in (a.len()..n).enumerate() {
-        a.insert(counter, '0');
+    let padding = n.saturating_sub(a.len());
+    if padding > 0 {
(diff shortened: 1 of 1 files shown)


src/math/karatsuba_multiplication.rs
```diff
@@ -35,8 +35,9 @@ fn _multiply(num1: i128, num2: i128) -> i128 {
 }
 
 fn normalize(mut a: String, n: usize) -> String {
-    for (counter, _) in (a.len()..n).enumerate() {
-        a.insert(counter, '0');
+    let padding = n.saturating_sub(a.len());
+    if padding > 0 {
+        a.insert_str(0, &"0".repeat(padding));
     }
```
vil02, line 41: What do you think about:
```suggestion
    a.insert_str(0, &"0".repeat(padding));
```

vil02 (changes requested):

src/math/karatsuba_multiplication.rs
dxdw233: Thanks! This looks good to me, just realized the if condition isn't necessary.

vil02 (approved):

PR #882, opened 2025-05-23
Add euler totient function

triuyen:
## Description
This PR adds an implementation of Euler's totient function φ(n) to the number theory section.
- Calculates φ(n) = count of integers ≤ n that are coprime to n (share no common factors except 1)
- Uses prime factorization method for efficient computation
- Complements the existing `compute_totient` function by providing single-value calculation

## Type of change
- **Input**: Single number `n` (vs. computing all values up to `n`)
- **Output**: Single φ(n) value (vs. vector of all values)
- **Use case**: Optimized for calculating φ(n) for one specific number, especially large numbers
- **Algorithm**: Prime factorization approach with O(√n) complexity

### Implementation details
- Function signature: `pub fn eulers_totient(n: u64) -> u64`
- Handles edge cases (n = 1)
- Uses integer arithmetic throughout (no floating point)
- Efficient prime factorization with early termination

## Checklist:

- [ Y ] I ran bellow commands using the latest version of **rust nightly**.
- [ Y ] I ran `cargo clippy --all -- -D warnings` just before my last commit and fixed any issue that was found.
- [ Y ] I ran `cargo fmt` just before my last commit.
- [ Y ] I ran `cargo test` just before my last commit and all tests passed.
- [ Y ] I added my algorithm to the corresponding `mod.rs` file within its own folder, and in any parent folder(s).
- [ Y ] I added my algorithm to `DIRECTORY.md` with the correct link.
- [ Y ] I checked `COUNTRIBUTING.md` and my code follows its guidelines.


vil02:
I am having a hard time with this one, cf. https://github.com/TheAlgorithms/Rust/pull/875#pullrequestreview-2696400035.

@siriak: what is your opinion?

@triuyen: in case it will be decided that it will be merged: please reach the _100% coverage_ and express the tests in a more _parametrized_ way.

siriak (commented):
The code looks good, I think we can still include this one because it's more efficient, it doesn't allocate all divisors in array to get their number. Just make sure to parameterize tests, thanks!

triuyen:
Thanks! I’ll go ahead and parameterize the tests as suggested. Let me know if anything else is needed before merging.

.gitignore
vil02, line 1: Why are you changing this file?

vil02:
@triuyen: regarding the failing clippy: #883 should resolve that issue. As soon as it will be merged, please update your branch.

.gitignore
triuyen: i wanted to give you guys a report of the 100 % coverage. i can put it back as it was in that case. Sorry

triuyen:
@vil02 Okay Thanks  for the info about #883 . I will update my branch after the merge

src/number_theory/euler_totient.rs
```diff
@@ -0,0 +1,106 @@
+pub fn euler_totient(n: u64) -> u64 {
+    if n == 1 {
+        return 1;
+    }
+
```
vil02, line 5: This appears to be redundant.
```suggestion
```

src/number_theory/euler_totient.rs
```diff
@@ -0,0 +1,106 @@
...
+    #[test]
+    fn test_prime_property() {
+        // For any prime p, φ(p) = p - 1
+        let primes = vec![2, 3, 5, 7, 11, 13, 17, 19, 23, 29, 31];
+        for p in primes {
+            assert_eq!(euler_totient(p), p - 1, "φ({p}) should be {}", p - 1);
+        }
+    }
```
vil02, line 105: What do you think about expressing these tests as:
```suggestion
    macro_rules! test_euler_totient {
        ($($name:ident: $test_case:expr,)*) => {
            $(
                #[test]
                fn $name() {
                    let (input, expected) = $test_case;
                    assert_eq!(euler_totient(input), expected)
                }
            )*
        };
    }

    test_euler_totient! {
        prime_2: (2, 1),
        prime_3: (3, 2),
        prime_5: (5, 4),
        prime_7: (7, 6),
        // ...
        composite_10: (10, 4), // 2 * 5
        composite_15: (15, 8), // 3 * 5
        // ...
        prime_power_2_to_3: (8, 4),
        prime_power_3_to_2: (9, 6),
        // ...
    }
```

The advantage is that if one case will fail, its name will appear in the output.

Clearly, please add the remaining cases.

vil02 (changes requested):

vil02 (approved):

PR #896, opened 2025-07-04
Triangular number

WinterPancake:
# Pull Request Template

## Description

A triangular number is a figurate number that counts objects arranged in an equilateral triangle. The standard formula for finding these can result in integer overflow due to the multiplication, so I added a OR binary operation to be able to find larger integers.
https://en.wikipedia.org/wiki/Triangular_number

## Type of change

## Checklist:


src/math/triangular_number.rs
```diff
@@ -0,0 +1,30 @@
+// Triangular Numbers: Function to the Nth Triangular Number
+// Wikipedia Reference : https://en.wikipedia.org/wiki/Triangular_number
```
vil02, line 2: Please explain what this module provides.

src/math/triangular_number.rs
```diff
@@ -0,0 +1,30 @@
...
+use num_bigint::BigUint;
+pub fn triangular_number(n: u64) -> BigUint {
```
vil02, line 4: `BigUint` seems to be an overkill here.

src/math/triangular_number.rs
```diff
@@ -0,0 +1,30 @@
...
+    let m: BigUint = ((n | 1) * ((n + 1) / 2)).into();
+    m
```
vil02, line 6: Why not:
```suggestion
    ((n | 1) * ((n + 1) / 2)).into()
```

src/math/triangular_number.rs
```diff
@@ -0,0 +1,30 @@
...
+                assert_eq!(triangular_number(n), expected);
+            }
+        )*
+        }
+    }
+
+    test_triangular_number! {
+        input_5: (6, BigUint::from(21u32)),
+        input_9: (10, BigUint::from(55u32)),
+        input_10: (100, BigUint::from(5050u32)),
```
vil02, line 28: What is the logic behind the names of the test cases?

Please add a text case with _odd_ input and the _first ones_ like:
```rust
        input_0: (0, BigUint::from(0u32)),
        input_1: (1, BigUint::from(1u32)),
```

vil02 (changes requested):

src/math/triangular_number.rs
WinterPancake: Added explanation to latest commit.

PR #903, opened 2025-08-18
Add black scholes algo

triuyen:
## Description
Black Sholes Algorithm :
This is an implementation of the Black-Scholes option pricing model, a mathematical formula used in finance to calculate the theoretical price of call options.

Calculates d1 and d2 values (statistical measures)
Uses normal distribution (normal_cdf) to find probabilities
Combines these to get the option's fair value

**### Also debug a lot of other codes due to recent change of clippy warnings. Please re-check to make sure.**
## Type of change

- [V] Bug fix (non-breaking change which fixes an issue)
- [V] New feature (non-breaking change which adds functionality)
- [V] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist:

- [V] I ran bellow commands using the latest version of **rust nightly**.
- [V] I ran `cargo clippy --all -- -D warnings` just before my last commit and fixed any issue that was found.
- [V] I ran `cargo fmt` just before my last commit.
- [V] I ran `cargo test` just before my last commit and all tests passed.
- [V] I added my algorithm to the corresponding `mod.rs` file within its own folder, and in any parent folder(s).
- [V] I added my algorithm to `DIRECTORY.md` with the correct link.
- [V] I checked `COUNTRIBUTING.md` and my code follows its guidelines.

### DIFF
--- a/src/data_structures/probabilistic/bloom_filter.rs
+++ b/src/data_structures/probabilistic/bloom_filter.rs
@@ -23,7 +23,6 @@ pub trait BloomFilter<Item: Hash> {
 /// If it's `true` the item may be present, or maybe another one produces the same hash
 #[allow(dead_code)]
 #[derive(Debug)]
-#[allow(dead_code)]
 struct BasicBloomFilter<const CAPACITY: usize> {
     vec: [bool; CAPACITY],
 }


triuyen:
Hello @imp2002, @vil02 !

I hope you’re both doing well !

I just wanted to kindly check in regarding my current pull request. 🙂
No urgency at all — I’d just love to hear your thoughts whenever you get a chance.

vil02:
@triuyen: please update your branch and make your changes _minimal_ (i.e. to have only your implementation of the added algorithm).

triuyen:
Okay ! Thank-you for the feedback @vil02 . 
For all the errors from other modules from running :  cargo clippy --all -- -D warnings , do i ignore those ?  
My branch is uptodate with the lastest change and i use the latest version of rust nightly
There is about : 21 errors on multiple modules

PR #912, opened 2025-09-23
fix(huffman): Handle edge cases and improve error handling

jesper-olsen:
## Description

This PR improves the robustness of the Huffman encoding implementation by properly handling edge cases that previously caused undefined behavior or panics.

### Summary of changes:
- Changed `HuffmanDictionary::new()` to return `Option<Self>` for safer API design
- Added proper handling for empty alphabets (returns `None` instead of panicking)
- Added special case handling for single-symbol alphabets, which previously failed during tree construction
- Improved error handling in `decode()` by replacing `unwrap()` calls with `?` operator
- Added `#[inline(always)]` optimization for the frequently-called `get_bit()` method
- Added comprehensive documentation with usage examples
- Added tests for edge cases (empty and single-symbol inputs)

The Huffman encoding algorithm is a well-known lossless data compression algorithm. More details: https://en.wikipedia.org/wiki/Huffman_coding

**Note**: During development, I observed intermittent heap corruption errors in the test suite that occur even without these changes, suggesting pre-existing issues in the codebase.

## Type of change

## Checklist:

### DIFF
--- a/src/machine_learning/loss_function/kl_divergence_loss.rs
+++ b/src/machine_learning/loss_function/kl_divergence_loss.rs
@@ -16,7 +16,7 @@ pub fn kld_loss(actual: &[f64], predicted: &[f64]) -> f64 {
     let loss: f64 = actual
         .iter()
         .zip(predicted.iter())
-        .map(|(&a, &p)| (a + eps) * ((a + eps) / (p + eps)).ln())
+        .map(|(&a, &p)| ((a + eps) * ((a + eps) / (p + eps)).ln()))
         .sum();
     loss
 }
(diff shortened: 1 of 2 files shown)


src/machine_learning/loss_function/kl_divergence_loss.rs
```diff
@@ -16,7 +16,7 @@ pub fn kld_loss(actual: &[f64], predicted: &[f64]) -> f64 {
     let loss: f64 = actual
         .iter()
         .zip(predicted.iter())
-        .map(|(&a, &p)| ((a + eps) * ((a + eps) / (p + eps)).ln()))
+        .map(|(&a, &p)| (a + eps) * ((a + eps) / (p + eps)).ln())
```
vil02, line 19: This seems not relevant. Please revert it here and create another tiny PR - we will merge it.

vil02 (changes requested):
The change looks good. One minor thing below.

PR #947, opened 2025-11-12
fix: prevent null pointer dereference in rb_tree delete_fixup

zhangzhuang15:
# Pull Request Template

## Description
Delete rb-tree node and rebalance up to root node, this will bring null pointer dereference at  start of loop  in delete_fixup function. To solve this problem, I add a guard and a test case.

## Type of change
BugFix

## Checklist:


zhangzhuang15:
@imp2002 cargo clippy, this checkpoint, should be improved! It complains code error in many files, but I don't cover these files in my PR.

PR #1030, opened 2026-03-24
feat: add SHA-1 hash function implementation

AliAlimohammadi:
## Description

Adds a pure-Rust implementation of the **SHA-1** cryptographic hash function, producing a 160-bit (20-byte) digest from an arbitrary-length byte slice.

SHA-1 is specified in [NIST FIPS 180-4](https://csrc.nist.gov/publications/detail/fips/180/4/final). The implementation follows the spec directly:

1. **Padding** — appends `0x80`, zero bytes, and the 64-bit big-endian message length so the total length is a multiple of 512 bits.
2. **Message schedule** — expands 16 big-endian words per block to 80 via `ROTL₁(wₜ₋₃ ⊕ wₜ₋₈ ⊕ wₜ₋₁₄ ⊕ wₜ₋₁₆)`.
3. **Compression** — 80 rounds across four 20-round stages, each with its own mixing function (`Ch`, `Parity`, `Maj`) and round constant.
4. **Serialisation** — five 32-bit state words written back as big-endian bytes.

> ⚠️ SHA-1 is cryptographically broken and should not be used for security-sensitive purposes. This implementation is provided for educational and algorithmic completeness only.

## Type of change

## Checklist


AliAlimohammadi:

AliAlimohammadi:

PR #1034, opened 2026-04-08
feat: add Fletcher checksum to hashing

AliAlimohammadi:
## Description

Adds an implementation of the **Fletcher-16 checksum** algorithm to `src/hashing/`.

Fletcher's checksum is a position-dependent checksum devised by John G. Fletcher (1934–2012) at Lawrence Livermore Labs in the late 1970s. It provides error-detection properties approaching those of a CRC but with lower computational effort, by maintaining two running sums (`sum1`, `sum2`) each reduced modulo 255, then packing them into a single 16-bit value.

Reference: [Fletcher's Checksum — Wikipedia](https://en.wikipedia.org/wiki/Fletcher%27s_checksum)

## Type of change

## Checklist


AliAlimohammadi:

siriak (approved):
Looks good, thanks!

PR #1036, opened 2026-04-16
feat: add Strand Sort and fix `sort_utils` unused import warnings

AliAlimohammadi:
## Description

Implements [Strand Sort](https://en.wikipedia.org/wiki/Strand_sort), a comparison-based sorting algorithm that works by repeatedly extracting increasing subsequences ("strands") from the input and merging them into a growing result list.

**Algorithm:**
1. Remove the first element of the remaining input and start a new strand.
2. Scan the rest of the input left-to-right; whenever an element is ≥ the last element of the strand, pull it out and append it to the strand. One full pass yields one sorted strand.
3. Merge the strand into the accumulated result via a standard two-way merge.
4. Repeat until the input is empty.

Also fixes pre-existing unused import warnings in `sort_utils.rs` by gating `use rand::RngExt` and `use std::time::Instant` with `#[cfg(test)]`, matching the functions that use them.

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue) — unused import warnings in `sort_utils.rs`

## Checklist

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -9,7 +9,7 @@ nalgebra = "0.34.0"
 ndarray = "0.17.2"
 num-bigint = { version = "0.4", optional = true }
 num-traits = { version = "0.2", optional = true }
-rand = "0.10"
+rand = "0.10.1"
 
 [dev-dependencies]
 quickcheck = "1.0"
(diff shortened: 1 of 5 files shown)


AliAlimohammadi:

siriak (approved):
Looks good, thanks!

PR #1037, opened 2026-04-23
feat: Add decoupled weight decay (AdamW) to Adam optimizer

AliAlimohammadi:
## Description

Extends the existing `Adam` optimizer in `src/machine_learning/optimization/adam.rs` to support decoupled weight decay (AdamW), as introduced in [Decoupled Weight Decay Regularization](https://arxiv.org/abs/1711.05101) (Loshchilov & Hutter, 2019).

Rather than adding a separate `AdamW` struct, a single `weight_decay: f64` field (defaulting to `0.0`) is added to the existing `Adam` struct. When `weight_decay` is `0.0` the update is identical to standard Adam. When positive, the decay term $\lambda \cdot \theta_{t-1}$ is subtracted directly from the parameters after the adaptive gradient step, keeping it independent of the second moment scaling — the key correction AdamW makes over naive L2 regularisation inside Adam.

The `step` signature changes from `step(&mut self, gradients: &[f64])` to `step(&mut self, gradients: &[f64], params: &[f64])`, since the decoupled decay term requires the current parameter values. All existing tests are updated accordingly (zero-initialised params preserve the original expected values).

### Algorithm

Both variants share the same moment update rules:

$$m_t = \beta_1 m_{t-1} + (1 - \beta_1) g_t$$

$$v_t = \beta_2 v_{t-1} + (1 - \beta_2) g_t^2$$

$$\hat{m}_t = \frac{m_t}{1 - \beta_1^t}, \qquad \hat{v}_t = \frac{v_t}{1 - \beta_2^t}$$

They differ only in the parameter update step:

**Adam** — weight decay is absent (or equivalently, folded into $g_t$ as L2 regularisation, where it gets scaled down by $1/\sqrt{\hat{v}_t}$):

$$\theta_t = \theta_{t-1} - \alpha \cdot \frac{\hat{m}_t}{\sqrt{\hat{v}_t} + \epsilon}$$

**AdamW** — weight decay is applied directly to $\theta_{t-1}$, fully decoupled from the adaptive scaling so its effect is constant regardless of gradient history:

$$\theta_t = \theta_{t-1} - \alpha \cdot \frac{\hat{m}_t}{\sqrt{\hat{v}_t} + \epsilon} - \alpha \lambda \theta_{t-1}$$

Setting $\lambda = 0$ in the AdamW update recovers Adam exactly, both mathematically and in the implementation (verified by
`test_adamw_step_weight_decay_zero_matches_adam`).

## Type of change

> `Adam::step` gains a required `params: &[f64]` argument. Existing callers must pass the current parameter slice; passing `&vec![0.0; n]` restores the original behaviour exactly.

## Checklist

### DIFF
--- a/src/machine_learning/optimization/adam.rs
+++ b/src/machine_learning/optimization/adam.rs
@@ -5,12 +5,19 @@
 //! learning problems. Boasting memory-efficient fast convergence rates, it sets and iteratively
 //! updates learning rates individually for each model parameter based on the gradient history.
 //!
+//! Setting `weight_decay > 0.0` enables the AdamW variant (Loshchilov & Hutter, 2019), which
+//! applies weight decay directly to the parameters rather than folding it into the gradients.
+//! This keeps the decay rate constant and independent of the gradient history — the key flaw
+//! that AdamW corrects over naive L2 regularization inside Adam. With `weight_decay = 0.0`
+//! (the default), the two algorithms are identical.
+//!
 //! ## Algorithm:
 //!
 //! Given:
 //!   - α is the learning rate
 //!   - (β_1, β_2) are the exponential decay rates for moment estimates
 //!   - ϵ is any small value to prevent division by zero
+//!   - λ is the weight decay coefficient (0.0 for standard Adam, > 0.0 for AdamW)
 //!   - g_t are the gradients at time step t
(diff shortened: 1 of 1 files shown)


AliAlimohammadi:

siriak (approved):
Looks good, thanks!

PR #1043, opened 2026-06-02
style/format-code: reduce code nesting

Tellcts:
## Description

1.  Reduced the multiple layers of nesting in match, making the code appear flatter;
2.  Remove the explicit import of Deref and fully leverage the compiler's implicit dereferencing behavior;  
3.  Optimize code logic.

## Type of change
- [x] Optimization of code structure

## Checklist:


Tellcts (commented):
I have review my codes and impove the test coverage.

PR #1063, opened 2026-08-05
fix: empty slice panics in bead_sort, bogo_sort, comb_sort and wave_sort

SEPURI-SAI-KRISHNA:
# Pull Request Template

## Description

Four algorithms in `src/sorting/` panic when handed an empty slice. Sorting nothing should
be a no-op, and the other 30 sorts in the module already treat it that way — 20 of them have
an explicit `empty()` test. These four were simply never exercised on that input.

```rust
sorting::bead_sort(&mut []);                 // index out of bounds: the len is 0 but the index is 0
sorting::bogo_sort(&mut [] as &mut [i32]);   // attempt to subtract with overflow
sorting::comb_sort(&mut [] as &mut [i32]);   // attempt to subtract with overflow
sorting::wave_sort(&mut [] as &mut [i32]);   // attempt to subtract with overflow
```

Each is an unsigned underflow or an unchecked first-element read:

* **`bead_sort`** — `let mut max = a[0];` reads element `0` before checking the slice is
  non-empty. Now returns early; there is no bead grid to build.
* **`bogo_sort`** — the private `is_sorted` helper loops `for i in 0..len - 1`, which
  underflows to `usize::MAX` at `len == 0`. Rewritten with `windows(2)`, which naturally
  yields nothing below two elements. This also drops a hand-rolled loop in favour of the
  same idiom `sorting::is_sorted` already uses.
* **`comb_sort`** — `gap` is clamped to a minimum of `1`, so `for i in 0..arr.len() - gap`
  underflows at `arr.len() == 0`. Now returns early for fewer than two elements.
* **`wave_sort`** — `for i in (0..n - 1).step_by(2)` underflows at `n == 0`. Now uses
  `n.saturating_sub(1)`.

These are release-mode hazards as well as debug-mode panics: with overflow checks off, the
underflowed bound becomes a near-`usize::MAX` loop bound and the sort runs off the end of
the slice.

### Tests

`empty` and `one_element` cases added to all four, following the module's existing
convention of asserting with the shared `is_sorted` / `have_same_elements` helpers.
`bead_sort` also gains an `all_zeroes` case, since a maximum of `0` is the other way its
bead grid can end up with zero columns. All new tests run in well under 300ms.

## Type of change

## Checklist:

> Note: no `mod.rs` or `DIRECTORY.md` change was needed — this fixes existing entries rather
> than adding a new algorithm.

### DIFF
--- a/src/sorting/bead_sort.rs
+++ b/src/sorting/bead_sort.rs
@@ -1,6 +1,11 @@
 //Bead sort only works for sequences of non-negative integers.
 //https://en.wikipedia.org/wiki/Bead_sort
 pub fn bead_sort(a: &mut [usize]) {
+    // An empty slice has no maximum to seed the bead grid with.
+    if a.is_empty() {
+        return;
+    }
+
     // Find the maximum element
     let mut max = a[0];
     (1..a.len()).for_each(|i| {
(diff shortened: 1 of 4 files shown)


SEPURI-SAI-KRISHNA:

Recapping the impact: four sorts panic on an empty slice where the other 30 in the module no-op. `bead_sort` reads `a[0]` before checking the slice is non-empty; the other three underflow an unsigned loop bound (`0..len - 1`, `0..arr.len() - gap`, `0..n - 1`). Worth noting these are not only debug-mode panics, with overflow checks disabled the bound becomes a near-`usize::MAX` value and the sort runs off the end of the slice.

